PR #5, opened 2021-06-22
metadata: handle network status as versioned Result

dcbw:
Instead of assuming the status will always be a
certain version, let CNI detect the version from
CNIVersion in the Result, and create the correct
concrete type for it.

When getting IP and MAC from a Result, convert to
a concrete type before pulling the info out.


dcbw:
@squeed @mrunalp

adrianreber:
Thanks for the patch. I see you changed the return result of the functions? Any specific reason? Just worried that this might require a new major release when following semver.

dcbw:
@adrianreber yes, it probably would require that. The problem with the existing code is that it returned a concrete versioned CNI Result type rather than the Result *interface* that all concrete versioned Result types implement. This forces checkpoint-restore and consumers of checkpoint-restore (CRIO, etc) to continuously upgrade their code and imports every time a new CNI spec version is created (and thus potentially a new concrete versioned CNI Result struct).

I can revert that part if you like.

The core problem is that ReadContainerCheckpointNetworkStatus() attempts to read JSON directly into a specific concrete versioned CNI Result type without ensuring that the version in that JSON actually matches the concrete type it's being read into. CNI provides version converters for that purpose that correctly up and down convert between different CNI spec types, so for max compatibility we should use those.

Best practice is to always use the Result interface, and convert to a concrete versioned Result when you actually need to manipulate or read the data. Otherwise it can just be a black box for saving/restoring.

adrianreber:
Sounds reasonable. You are referencing the 1.0.0-rc1 release in this PR. Would waiting for the final 1.0.0 release make sense to you before merging this?

dcbw:
> Sounds reasonable. You are referencing the 1.0.0-rc1 release in this PR. Would waiting for the final 1.0.0 release make sense to you before merging this?

@adrianreber sure, that is reasonable

adrianreber:
Closing as CNI dependency has been removed.

PR #7, opened 2021-09-17
Remove cni dependency

### LINKED ISSUE #6: Do not use the podman network status
Podman is going to abstract CNI from libpod (see https://github.com/containers/podman/pull/11232).
When the work is done libpod will no longer use the CNI result as network status. I am currently wiring the new interface into libpod and noticed that the CNI result is dumped into the checkpoint dir and that this library is using the result on restore to the set the ip and mac address. In my opinion this library should not have to know anything about the network status used in podman. I think it would make more sense to handle setting the correct ip and mac entirely in libpod and not here.

Luap99:
Podman 4.0 no longer uses CNI to store the network information therefore
reading the status file into a cni result will fail. This library should
not make any assumptions about the network status format used by Podman.
I personally don't see much value in displaying the IP address with
checkpointctl for a podman container. A container can have 0-n IP
addresses. Which one should be displayed anyway?

Fixes #6

Signed-off-by: Paul Holzinger <pholzing@redhat.com>


rhatdan:
@adrianreber PTAL

adrianreber:
I am still not sure if this would not be necessary if the proposed fixes from #5 are merged.

I was also concerned how this works with restoring IP addresses during container restore, but I just saw that you changed all that code a month ago. Has the format of the file `network.status` changed in Podman because of your changes to Podman? Is it no  longer possible to restore a container checkpointed with the old version of `network.status`?

Luap99:
> I am still not sure if this would not be necessary if the proposed fixes from #5 are merged.

Well #5 would be fine for podman because then I could bump the CNI version in podman. However there is no longer a reason for this lib to include CNI because the network status in podman no longer uses CNI structures. 

> I was also concerned how this works with restoring IP addresses during container restore, but I just saw that you changed all that code a month ago. Has the format of the file `network.status` changed in Podman because of your changes to Podman?

Yes the format has changed, we no longer store a CNI result. We store a `map[string]StatusBlock` defined in https://github.com/containers/podman/blob/b906ecbb5bc887f5123a0f61a5a51782aa34357f/libpod/network/types/network.go#L120
While we could update this lib to use the new structure I personally see no reason why this is needed. I don't want to make any guarantees that this structure will remain stable. It is internal to libpod and can therefore be changed any time as long as we handle the migration inside libpod.

The only reason for this lib to use the network status is to show the IP address with `checkpointctl`.
A podman container can have 0-n IP addresses. Which one should be displayed anyway? I would prefer `checkpointctl` to not show IP addresses for podman containers. 

> Is it no longer possible to restore a container checkpointed with the old version of `network.status`?

Since this change is for podman 4.0 I see no problems to break old checkpoints, only the network information is lost. Besides the network settings the restore of an old 3.X checkpoint should still work correctly.

adrianreber:
> > I am still not sure if this would not be necessary if the proposed fixes from #5 are merged.
> 
> Well #5 would be fine for podman because then I could bump the CNI version in podman. However there is no longer a reason for this lib to include CNI because the network status in podman no longer uses CNI structures.

The reason why I created checkpointctl and the library is that I want to use the same code in CRI-O, Podman and Kubernetes as far as possible. As I am able to move checkpoints from Podman to CRI-O it is important to have compatible checkpoint archives.

> It is internal to libpod and can therefore be changed any time as long as we handle the migration inside libpod.

Would be nice to have a format which is independent of internal changes.

But currently in my CRI-O PR I am not using the network part of checkpointctl, so the code can be dumped and later introduced if necessary. But having a format for network information which is independent of an internal struct would be nice.

> > Is it no longer possible to restore a container checkpointed with the old version of network.status?
>
> Since this change is for podman 4.0 I see no problems to break old checkpoints, only the network information is lost. Besides the network settings the restore of an old 3.X checkpoint should still work correctly.

That does not sound like the best solution.

Anyway, if you can make CI happy, this can be merged.

Luap99:
Podman is actively moving away from CNI so using the same code for networking will be difficult.

If you want to know about network settings with a stable format I would recommend to define a struct here in this lib with the fields you would like to know about. If there is such a struct the consumers (podman, crio, ...) would be required to fill in the data and dump in into the network status file. With that this lib could read it without having to worry about changes to the format.

As long as podman can still dump the raw format into the checkpoint to read advanced options I would be fine with using a fixed struct from this lib for the network status file.

PR #15, opened 2022-03-18
metadata: Add CheckpointVolumesDirectory

rst0git:
The `volumes` directory was introduced in Podman with https://github.com/containers/podman/pull/8781. It is used to store the content of volumes associates with checkpointed container.


rst0git:
@adrianreber the `volumes` field is used in https://github.com/containers/podman/pull/13505

PR #16, opened 2022-08-17
Upgrade go-criu to v6

snprajwal:
This commit replaces the deprecated `stats` package with `crit`, in
accordance with the change in the new release of go-criu.


adrianreber:
Before merging this I want to get two other PRs merged first which would reduce the amount of required changes in this PR a bit as I want to drop unneeded functionality.

adrianreber:
Please rebase now.

container.go
```diff
@@ -154,11 +155,14 @@ func showContainerCheckpoint(checkpointDirectory string) error {
 	}
 	defer cpDir.Close()
 
-	dumpStatistics, err := stats.CriuGetDumpStats(cpDir)
+	// Get dump statistics with crit
+	c := crit.New(filepath.Join(cpDir.Name(), "stats-dump"), "", "", false, false)
+	statsImg, err := c.Decode()
```
adrianreber, line 160: Can we provide a function in go-criu to do exactly this. I see you are also using the same code in Podman. So maybe it would be good to have it available in go-criu.

snprajwal: Leaving it open-ended would mean that the user can do this with any of the images, not just `stats-dump` and `stats-restore`. `crit` provides a generalised way to operate on all images. I feel it would be better to let the user deal with the images, rather than having specific functions for any one type. What do you think?

rst0git: Thank you Adrian for your input. I assume that you are referring to the following change in Podman:
```
```

rst0git: Thank you Adrian for your input! I assume that you are referring to the following change in Podman.
```diff
-		dumpStatistics, err := stats.CriuGetDumpStats(statsDirectory)
+		// Get dump statistics with crit
+		c := crit.New(filepath.Join(statsDirectory.Name(), "stats-dump"), "", "", false, false)
+		statsImg, err := c.Decode()
```
Indeed, `CriuGetDumpStats()` looks straightforward and easier to use, whereas crit provides more flexibility.
@snprajwal Perhaps we could add `CriuGetDumpStats()` as a helper function in go-criu?

snprajwal: Will get this pushed next week, we can publish it on v6.1.0 and use it

rst0git (approved):
LGTM

PR #23, opened 2022-09-01
Removed unnecessary code for pod checkpoints

adrianreber:
Initially checkpointctl had support to handle checkpoints from pods with multiple containers as implemented in a proof of concept for CRI-O as well as support for multiple Kubernetes checkpoints which were also part of a proof of concept. As all this code was never published besides as a proof of concept in a pull request, this code has now been removed.


rst0git (approved):
LGTM

PR #41, opened 2023-03-17
make: Add install and uninstall targets

rst0git:
This pull request adds install and uninstall Makefile targets for the checkpointctl tool.


adrianreber:
Not sure, but would a make install and make uninstall during test make sense?

rst0git:
I've updated the pull request with a test for the install/uninstall make targets and documentation for the steps to install checkpointctl from source code.

PR #42, opened 2023-03-17
Add checkpointctl version

rst0git:
This pull request adds `--version` option for the checkpointctl tool.


adrianreber:
What about passing the version from the Makefile. Search for main.name for an example.

rst0git:
Thanks @adrianreber! I've updated the pull request to pass the version from the Makefile using `main.version`.

PR #43, opened 2023-04-13
Add '--mounts' option

sankalp-12:
The `checkpointctl` tool could be extended  to display an overview of mounts used by the checkpoints. 
This pull request adds  `--mounts` option for the `checkpointctl show` command, which would extend the *out* to look like this:

```
checkpointctl show checkpoint.tar.gz --mounts

Displaying container checkpoint data from /tmp/checkpointctl1209302006

+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
|   CONTAINER    |             IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| cool_heyrovsky | docker.io/library/httpd:latest | 7fb253cbd7c0 | crun    | 2023-04-14T04:46:00+05:30 | Podman | 5.3 MiB    | 2.0 KiB           |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+

Overview of Mounts
+--------------------+--------+---------------------------+
|    DESTINATION     |  TYPE  |          SOURCE           |
+--------------------+--------+---------------------------+
| /proc              | proc   | proc                      |
| /dev               | tmpfs  | tmpfs                     |
| /sys               | sysfs  | sysfs                     |
| /dev/pts           | devpts | devpts                    |
| /dev/mqueue        | mqueue | mqueue                    |
| /etc/hostname      | bind   | ../userdata/hostname      |
| /run/.containerenv | bind   | ../userdata/.containerenv |
| /etc/resolv.conf   | bind   | ../userdata/resolv.conf   |
| /etc/hosts         | bind   | ../userdata/hosts         |
| /dev/shm           | bind   | ../userdata/shm           |
| /sys/fs/cgroup     | cgroup | cgroup                    |
+--------------------+--------+---------------------------+
```
Also, a `--full-paths` option is added to display the mounts with the full paths. The *output* would look like this:

```
checkpointctl show checkpoint.tar.gz --mounts --full-paths

Displaying container checkpoint data from /tmp/checkpointctl2681949126

+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
|   CONTAINER    |             IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| cool_heyrovsky | docker.io/library/httpd:latest | 7fb253cbd7c0 | crun    | 2023-04-14T04:46:00+05:30 | Podman | 5.3 MiB    | 2.0 KiB           |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+

Overview of Mounts
+--------------------+--------+------------------------------------------------------------------------------------------------------------------------------------+
|    DESTINATION     |  TYPE  |                                                               SOURCE                                                               |
+--------------------+--------+------------------------------------------------------------------------------------------------------------------------------------+
| /proc              | proc   | proc                                                                                                                               |
| /dev               | tmpfs  | tmpfs                                                                                                                              |
| /sys               | sysfs  | sysfs                                                                                                                              |
| /dev/pts           | devpts | devpts                                                                                                                             |
| /dev/mqueue        | mqueue | mqueue                                                                                                                             |
| /etc/hostname      | bind   | /run/containers/storage/overlay-containers/7fb253cbd7c0c30dd2cc26adbefc2817fd3eb2ce48f8c3110ded8dac19a6bcfa/userdata/hostname      |
| /run/.containerenv | bind   | /run/containers/storage/overlay-containers/7fb253cbd7c0c30dd2cc26adbefc2817fd3eb2ce48f8c3110ded8dac19a6bcfa/userdata/.containerenv |
| /etc/resolv.conf   | bind   | /run/containers/storage/overlay-containers/7fb253cbd7c0c30dd2cc26adbefc2817fd3eb2ce48f8c3110ded8dac19a6bcfa/userdata/resolv.conf   |
| /etc/hosts         | bind   | /run/containers/storage/overlay-containers/7fb253cbd7c0c30dd2cc26adbefc2817fd3eb2ce48f8c3110ded8dac19a6bcfa/userdata/hosts         |
| /dev/shm           | bind   | /var/lib/containers/storage/overlay-containers/7fb253cbd7c0c30dd2cc26adbefc2817fd3eb2ce48f8c3110ded8dac19a6bcfa/userdata/shm       |
| /sys/fs/cgroup     | cgroup | cgroup                                                                                                                             |
+--------------------+--------+------------------------------------------------------------------------------------------------------------------------------------+

```

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -14,6 +14,7 @@ var (
 	name       string
 	version    string
 	printStats bool
+	showMounts bool
 )
 
 func main() {
@@ -47,6 +48,12 @@ func setupShow() *cobra.Command {
 		"print-stats",
 		false,
 		"Print checkpointing statistics if available",
+	)
+	flags.BoolVar(
+		&showMounts,
+		"mounts",
+		false,
+		"Print overview about mounts used in the checkpoints",
 	)
 
 	return cmd
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -15,6 +15,7 @@ var (
 	version    string
 	printStats bool
 	showMounts bool
+	fullPaths  bool
 )
 
 func main() {
@@ -55,11 +56,21 @@ func setupShow() *cobra.Command {
 		false,
 		"Print overview about mounts used in the checkpoints",
 	)
+	flags.BoolVar(
+		&fullPaths,
+		"full-paths",
+		false,
+		"Display mounts with full paths",
+	)
 
 	return cmd
 }
 
 func show(cmd *cobra.Command, args []string) error {
+	if fullPaths && !showMounts {
+		return fmt.Errorf("Cannot use --full-paths without --mounts option")
+	}
+
 	input := args[0]
 	tar, err := os.Stat(input)
 	if err != nil {
@@ -81,6 +92,5 @@ func show(cmd *cobra.Command, args []string) error {
 	if err := archive.UntarPath(input, dir); err != nil {
 		return fmt.Errorf("unpacking of checkpoint archive %s failed: %w", input, err)
 	}
-
 	return showContainerCheckpoint(dir)
 }
--- a/container.go
+++ b/container.go
@@ -160,7 +160,12 @@ func showContainerCheckpoint(checkpointDirectory string) error {
 			table.Append([]string{
 				data.Destination,
 				data.Type,
-				shortenPath(data.Source),
+				func() string {
+					if fullPaths {
+						return data.Source
+					}
+					return shortenPath(data.Source)
+				}(),
 			})
 		}
 		fmt.Println("\nOverview of Mounts")
--- a/container.go
+++ b/container.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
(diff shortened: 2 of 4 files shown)


adrianreber:
Thanks for the PR. Can you add more details to the commit message (and maybe PR description) how the output would look like for example.

This would also need some tests and a happy CI.

sankalp-12:
> Thanks for the PR. Can you add more details to the commit message (and maybe PR description) how the output would look like for example.
> 
> This would also need some tests and a happy CI.

I have added some details on how the output would look in the PR description. 

AFAIU, the tests would be added in the `test/checkpointctl.bats` file. Looking through some of the previous tests written (like for `--print-stats` option), a populated `stats-dump` file is present as a CRIU image. Similarly, for testing the `--mounts` option, should I populate the `spec.dump` file?

adrianreber:
> I have added some details on how the output would look in the PR description.

Looks good. Please also add it to the commit so that it can be seen without GitHub.

> > Thanks for the PR. Can you add more details to the commit message (and maybe PR description) how the output would look like for example.
> > This would also need some tests and a happy CI.
> 
> I have added some details on how the output would look in the PR description.
> 
> AFAIU, the tests would be added in the `test/checkpointctl.bats` file. Looking through some of the previous tests written (like for `--print-stats` option), a populated `stats-dump` file is present as a CRIU image. Similarly, for testing the `--mounts` option, should I populate the `spec.dump` file?

Yes, adding one or two mounts to our test data sounds like the right thing to do.

Code coverage errors are being fixed in #44

@rst0git what do you think about the name `--mounts`. We have `--print-stats` and now `--mounts`? Should we prefix all those options with a common string like `--show` or `--print`? For `--print-stats` we can introduce an alias or something.

sankalp-12:
> Looks good. Please also add it to the commit so that it can be seen without GitHub.
> Yes, adding one or two mounts to our test data sounds like the right thing to do.
 
 Done.

rst0git:
> what do you think about the name --mounts. We have --print-stats and now --mounts? Should we prefix all those options with a common string like --show or --print? For --print-stats we can introduce an alias or something.

The `print` prefix looks unnecessary because this option is used with the `show` command.
IMHO, the first would be more intuitive:
```
sudo checkpointctl show --mounts /tmp/chkpt.tar.gz
```
vs
```
sudo checkpointctl show --print-mounts /tmp/chkpt.tar.gz
```
We could also add an option like `--all` that could show everything.

adrianreber:
Good point. Makes sense. Then we could add a alias for stats.

test/spec.dump
```diff
@@ -1,3 +1,897 @@
 {
-"annotations": {"io.container.manager": "libpod"}
+  "ociVersion": "1.0.2-dev",
```
adrianreber, line 2: I would prefer if you only add one or two mounts. Just the minimum we need for tests. Nothing else.

container.go
```diff
@@ -206,3 +224,11 @@ func getCheckpointSize(path string) (size int64, err error) {
 
 	return dirSize(dir)
 }
+
+func shortenPath(path string) string {
```
adrianreber, line 228: I am not sure how I feel about this. I understand why it might be nice for very long paths, but I think it should be optional. Maybe default to this output, but also provide a way to display the full paths.

sankalp-12:
> Good point. Makes sense. Then we could add a alias for stats.

Should the alias for `--print-stats` be added in this PR itself?

adrianreber:
> > Good point. Makes sense. Then we could add a alias for stats.
> 
> Should the alias for `--print-stats` and an `--all` option suggested by @rst0git be added in this PR itself?

No, let's do that in another PR.

checkpointctl.go
```diff
@@ -48,6 +50,18 @@ func setupShow() *cobra.Command {
 		false,
 		"Print checkpointing statistics if available",
 	)
+	flags.BoolVar(
+		&showMounts,
+		"mounts",
+		false,
+		"Print overview about mounts used in the checkpoints",
+	)
+	flags.BoolVar(
```
adrianreber, line 59: Please add a check that this can only be used in combination with `--mounts`.

adrianreber: Also include a test that verifies this.

test/spec.dump
```diff
@@ -1,3 +1,29 @@
...
+  "mounts": [
+    {
+      "destination": "/proc",
+      "type": "proc",
+      "source": "proc",
+      "options": [
```
adrianreber, line 8: As you are only using `destination`, `type` and `source`, it seems unnecessary to include options.

test/spec.dump
```diff
@@ -1,3 +1,29 @@
 {
-"annotations": {"io.container.manager": "libpod"}
+  "ociVersion": "1.0.2-dev",
```
adrianreber, line 2: This line is also not needed for out test setup.

test/spec.dump
```diff
@@ -1,3 +1,29 @@
...
+      "type": "proc",
+      "source": "proc",
+      "options": [
+        "nosuid",
+        "noexec",
+        "nodev"
+      ]
+    },
+    {
+      "destination": "/dev",
```
adrianreber, line 15: You should create an entry that can show the differences of `shortenPath()` during the tests.

container.go
```diff
@@ -147,41 +148,66 @@ func showContainerCheckpoint(checkpointDirectory string) error {
...
+	if showMounts {
+		table = tablewriter.NewWriter(os.Stdout)
+		table.SetHeader([]string{
+			"Destination",
+			"Type",
+			"Source",
+		})
+		// Get overview of mounts from spec.dump
+		for _, data := range specDump.Mounts {
+			if fullPaths {
```
adrianreber, line 160: This is probably just me, but I would do something like this:
```go
table.Append([]string{
    data.Destination,
    data.Type,
    func() string {
        if fullPaths {
            return data.Source
        }
        return shortenPath(data.Source)
    }()
})
```

adrianreber: Not tested, but probably close. I just like using closures.

container.go
```diff
@@ -147,41 +148,67 @@ func showContainerCheckpoint(checkpointDirectory string) error {
...
+					}
+					return shortenPath(data.Source)
+				}(),
+			})
+		}
+		fmt.Println("\nOverview of Mounts")
+		table.Render()
+	} else {
+		if fullPaths {
+			return fmt.Errorf("Cannot use --full-paths without --mounts option")
```
adrianreber, line 175: Thanks. Please do this check much earlier. I would do it in `checkpointctl.go` before opening any files. No need to unpack a file to detect later that we will not use it at all.

checkpointctl.go
```diff
@@ -74,6 +88,9 @@ func show(cmd *cobra.Command, args []string) error {
 	if err := archive.UntarPath(input, dir); err != nil {
 		return fmt.Errorf("unpacking of checkpoint archive %s failed: %w", input, err)
 	}
-
+	
+	if (fullPaths && !showMounts) {
```
adrianreber, line 92: Sorry, but almost. If you move it to the top of this function it will be before the first call to `Stat()` and `UntarPath()`. This way no disk access happens if we just need to exit here.

adrianreber:
Looks good to me. (In `checkpointctl.go` there is one line removed which seems like an unnecessary change, but we can ignore that for now.)

@rst0git What do you think?

container.go
```diff
@@ -160,7 +160,12 @@ func showContainerCheckpoint(checkpointDirectory string) error {
 			table.Append([]string{
 				data.Destination,
 				data.Type,
-				shortenPath(data.Source),
+				func() string {
+					if fullPaths {
+						return data.Source
+					}
+					return shortenPath(data.Source)
+				}(),
```
rst0git, line 168: Could we move the check for `fullPaths` before `table.Append`?
For example:
```Go
source := data.Source
if !fullPaths {
	source = shortenPath(source)
}
table.Append([]string{
	data.Destination,
	data.Type,
	source,
})
```

sankalp-12: So which one do we go ahead with? :confused:

rst0git: I'm sorry, I missed your comment. Let's leave as it is.

rst0git (approved):
LGTM

PR #45, opened 2023-04-17
Show test results in GitHub PRs

adrianreber:
As bats supports creating junit compatible output, this output can be used to show an overview about test results directly in GitHub PRs as a comment.


adrianreber:
@rst0git what do you think of this test report?

rst0git (approved):
Sounds like a good idea.

PR #46, opened 2023-04-18
Add alias and '--all' option to checkpointctl show

sankalp-12:
This pull request adds an alias for the `--print-stats` option, to be changed to `--stats` to maintain consistency among option names of `checkpointctl show`.

Also, an `-all` option is added to display all additional information about the available checkpoints, which would extend the *out* to look like: 

```
checkpointctl show checkpoint.tar.gz --all

Displaying container checkpoint data from /tmp/checkpointctl1209302006

+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
|   CONTAINER    |             IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| cool_heyrovsky | docker.io/library/httpd:latest | 7fb253cbd7c0 | crun    | 2023-04-14T04:46:00+05:30 | Podman | 5.3 MiB    | 2.0 KiB           |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+

Overview of Mounts
+--------------------+--------+---------------------------+
|    DESTINATION     |  TYPE  |          SOURCE           |
+--------------------+--------+---------------------------+
| /proc              | proc   | proc                      |
| /dev               | tmpfs  | tmpfs                     |
| /sys               | sysfs  | sysfs                     |
| /dev/pts           | devpts | devpts                    |
| /dev/mqueue        | mqueue | mqueue                    |
| /etc/hostname      | bind   | ../userdata/hostname      |
| /run/.containerenv | bind   | ../userdata/.containerenv |
| /etc/resolv.conf   | bind   | ../userdata/resolv.conf   |
| /etc/hosts         | bind   | ../userdata/hosts         |
| /dev/shm           | bind   | ../userdata/shm           |
| /sys/fs/cgroup     | cgroup | cgroup                    |
+--------------------+--------+---------------------------+

CRIU dump statistics
+---------------+-------------+--------------+---------------+---------------+---------------+
| FREEZING TIME | FROZEN TIME | MEMDUMP TIME | MEMWRITE TIME | PAGES SCANNED | PAGES WRITTEN |
+---------------+-------------+--------------+---------------+---------------+---------------+
| 5333 us       | 200318 us   | 111941 us    | 75612 us      |        239777 |         45512 |
+---------------+-------------+--------------+---------------+---------------+---------------+
```

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -27,16 +27,20 @@ func main() {
 		SilenceUsage: true,
 	}
 
-	showCommand := setupShow()
+	showCommand, err := setupShow()
+	if err != nil {
+		os.Exit(1)
+	}
+	
 	rootCommand.AddCommand(showCommand)
 	rootCommand.Version = version
 
-	if err := rootCommand.Execute(); err != nil {
+	if err = rootCommand.Execute(); err != nil {
 		os.Exit(1)
 	}
 }
 
-func setupShow() *cobra.Command {
+func setupShow() (*cobra.Command, error) {
 	cmd := &cobra.Command{
 		Use:   "show",
 		Short: "Show information about available checkpoints",
@@ -47,6 +51,12 @@ func setupShow() *cobra.Command {
 	flags.BoolVar(
 		&printStats,
 		"print-stats",
+		false,
+		"Print checkpointing statistics if available",
+	)
+	flags.BoolVar(
+		&printStats,
+		"stats",
 		false,
 		"Print checkpointing statistics if available",
 	)
@@ -63,7 +73,8 @@ func setupShow() *cobra.Command {
 		"Display mounts with full paths",
 	)
 
-	return cmd
+	err := flags.MarkHidden("print-stats")
+	return cmd, err
 }
 
 func show(cmd *cobra.Command, args []string) error {
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -4,6 +4,7 @@ package main
 
 import (
 	"fmt"
+	"log"
 	"os"
 
 	"github.com/containers/storage/pkg/archive"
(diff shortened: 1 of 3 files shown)


adrianreber:
You should not replace `--print-stats` but mark it as hidden. That way the old flag still works, but it is not shown in in the help output.

sankalp-12:
@adrianreber Would it be required to have tests with `--all` option with invalid/missing `*.dump` files?
Also, currently the default output of the `--all` option is with shortened paths. The user has an option to use `--full-paths` along with `--all`. Is this fine?

sankalp-12:
> You should not replace `--print-stats` but mark it as hidden. That way the old flag still works, but it is not shown in in the help output.

So, we mark `--print-stats` as hidden and introduce `--stats` as a new flag?

adrianreber:
> @adrianreber Would it be required to have tests with `--all` option with invalid/missing `*.dump` files?

Undecided. I would look at the existing code test coverage and decide using that data. If code is covered by tests then no need to introduce new tests. If it is not covered by tests, tests are needed.

> Also, currently the default output of the `--all` option is with shortened paths. The user has an option to use `--full-paths` along with `--all`. Is this fine?

Sounds good.

sankalp-12:
Shall I add a shorthand to all the options for *checkpointctl show*? Currently, only —help has one.

checkpointctl.go
```diff
@@ -62,13 +73,24 @@ func setupShow() *cobra.Command {
...
+	flags.BoolVar(
+		&showAll,
+		"all",
+		false,
+		"Display all additional information about the checkpoints",
+	)
 
-	return cmd
+	err := flags.MarkHidden("print-stats")
+	return cmd, err
```
snprajwal, line 84: Rather than refactoring the function to return an error, I feel it would be better if we do a `log.Fatal(err)` instead here.

snprajwal (approved):
LGTM

adrianreber:
> @adrianreber Shall I add a shorthand for all the options of `checkpointctl show`? Currently, only `—help` has one.

No real opinion on this. I would not add it for now. Maybe later.

adrianreber:
Looks good, indeed.

checkpointctl.go
adrianreber: I must say I am not the biggest friend for `Fatal()` :wink:. It doesn't go through the usual steps to exit the program. Not really a problem here, but I think returning an error is also something which makes sense. No need to change this again. Let's merge it as it is.

adrianreber:
Happy to see so good test coverage. Nice.

checkpointctl.go
snprajwal: O_o could you elaborate a little on the usual steps part? AFAIK, `log.Fatal` calls `os.Exit(1)` internally. Although I am not sure if it runs deferred functions before dying :thinking:

PR #55, opened 2023-05-25
Optimize archive unpacking

### LINKED ISSUE #51: Do not unpack the whole checkpoint archive
Currently checkpointctl is not really clever when it comes to showing information about a checkpoint archive. Checkpoint archives are always unpacked unconditionally and completely.

It would be enough, depending on the options used, to only unpack one or two files. This would require less space in `/tmp` and probably be faster.

The size of the checkpoint can also be figured out from the tar headers and does not require actual unpacking of the files.

behouba:
This PR propose an approach to optimize checkpoint archive unpacking. 

Two new functions are introduced:
- untarFiles` to extract specific files from the archive. 
- `getCheckpointAndDiffSizes`  as  replacement for `getCheckpointSize` function. This function read the sizes of the checkpoint directory and rootfs-diff.tar file from the archive without unpacking the entire archive.

Two failing test cases checking the presence of the `checkpoint` directory. Is it OK to remove those ?

cc: @adrianreber, @rst0git 

Fixes: #51

### DIFF
--- a/container.go
+++ b/container.go
@@ -5,8 +5,11 @@
 package main
 
 import (
+	"archive/tar"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -14,6 +17,7 @@ import (
 
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
 	"github.com/checkpoint-restore/go-criu/v6/crit"
+	"github.com/containers/storage/pkg/archive"
 	"github.com/olekukonko/tablewriter"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
 )
@@ -125,21 +129,18 @@ func showContainerCheckpoint(checkpointDirectory, input string) error {
 		row = append(row, ci.MAC)
 	}
 
-	size, err := getCheckpointSize(checkpointDirectory)
+	archiveSizes, err := getArchiveSizes(input)
 	if err != nil {
 		return err
 	}
 
 	header = append(header, "CHKPT Size")
-	row = append(row, metadata.ByteToString(size))
+	row = append(row, metadata.ByteToString(archiveSizes.checkpointSize))
 
 	// Display root fs diff size if available
-	fi, err := os.Lstat(filepath.Join(checkpointDirectory, metadata.RootFsDiffTar))
-	if err == nil {
-		if fi.Size() != 0 {
-			header = append(header, "Root Fs Diff Size")
-			row = append(row, metadata.ByteToString(fi.Size()))
-		}
+	if archiveSizes.rootFsDiffTarSize != 0 {
+		header = append(header, "Root Fs Diff Size")
+		row = append(row, metadata.ByteToString(archiveSizes.rootFsDiffTarSize))
 	}
 
 	table.SetAutoMergeCells(true)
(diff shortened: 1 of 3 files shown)


adrianreber:
Overall this looks very good.

> Two failing test cases checking the presence of the checkpoint directory. Is it OK to remove those ?

I would say you should adapt them. There should be an error message if the tar file has no `checkpoint` directory. Currently it seems to just says size is `0`. You probably should add an error message if there is no checkpoint directory.

Maybe have an early test to see if there is a  `checkpoint` directory in the tar file at all. If it is missing we should have some error message.

container.go
```diff
@@ -209,25 +209,48 @@ func showContainerCheckpoint(checkpointDirectory, input string) error {
 	return nil
 }
 
-func dirSize(path string) (size int64, err error) {
-	err = filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
-		if err != nil {
-			return err
+// getCheckpointAndDiffSizes retrieves the size of the checkpoint directory and
```
adrianreber, line 212: I think I understand your motivation. You loop once over the list of tar files and calculate both sizes.

I am not totally convinced this is the right way from an interface point of view. It feels like two things are returned from one function which do not belong together. At the same time it makes sense this way instead of looping twice over the tar archive.

The only idea I have is, is to have a function which does the same as you do currently, but then have two functions, one which returns the checkpoint size and the other returns the rootfs size. If a size is requested both sizes are calculated and stored in a variable and can then be access by the corresponding function. This sounds, however, like over-engineering...

@rst0git do you have an idea how to do this?

adrianreber:
Please also add new test cases for compressed tar achives to ensure that works and also add a rootfs size test.

container.go
behouba: What if we have a function that takes a list of string file patterns and returns the total size of the matching files (using the same logic)? This function could then be used within other functions to obtain the size of the `checkpoint` directory or the size of the rootfs tar file.

rst0git: maybe we could use something like the following:

```Go
type archiveSizes struct {
	checkpointSize    int64
	rootFsDiffTarSize int64
}

func GetArchiveSizes(archiveInput string) (*archiveSizes, error) {
	archiveFile, err := os.Open(archiveInput)
	if err != nil {
		return nil, err
	}
	defer archiveFile.Close()

	// Decompress the archive
	decompressedStream, err := archive.DecompressStream(archiveFile)
	if err != nil {
		return nil, err
	}
	defer decompressedStream.Close()

	// Create a tar reader to read the files from the decompressed archive
	tarReader := tar.NewReader(decompressedStream)

	// Initialize the singleton instance
	var result = &archiveSizes{}

	// Iterate through each file in the archive
	for {
		hdr, err := tarReader.Next()
		if errors.Is(err, io.EOF) {
			// End of archive
			break
		}
		if err != nil {
			break
		}

		if hdr.FileInfo().Mode().IsRegular() {
			if strings.HasPrefix(hdr.Name, metadata.CheckpointDirectory) {
				// Add the file size to the total checkpoint size
				result.checkpointSize += hdr.Size
			} else if hdr.Name == metadata.RootFsDiffTar {
				// Assuming there is only one rootfs diff tar file
				result.rootFsDiffTarSize = hdr.Size
			}
		}
	}

	return result, nil
}
```

adrianreber: Let's try what @rst0git suggested.

adrianreber:
Please take a look at why the test code coverage decreases. Maybe you need to create tests with broken tar files to handle a couple of your new error paths. Maybe just a compressed file without a tar archive or you can create a tar file and then replace parts of if to break it.

I would also prefer if you do not check in the tar file for the test case. You should be able to create it on the fly.

adrianreber:
Also, please do not create fixup commits. If you have fixes of your original commit then squash the fixes into the original commit of the PR.

To add the test cases in a separate commit makes sense from my point of view.

behouba:
Thank you so much for the valuable feedback @adrianreber 

I have modified the previous implementation be able to reuse the same logic were we need to iterate over archive header files. 
What do you think ?

Also, I am not sure if the test cases are good ?

adrianreber:
Thanks for the added tests. This looks really good know. I will need a couple of days to have a closer look.

rst0git (approved):
LGTM

PR #56, opened 2023-05-28
feat: add flag to view process tree

snprajwal:
CRIT already allows us to explore the process tree, file descriptors, memory map, and RSS map for a checkpoint. These features can be integrated with checkpointctl to allow more info to be displayed with `checkpointctl show`. This PR introduces `checkpointctl show --ps-tree` for viewing the process tree.

Here's a sample output of the tree displayed for a single process:

```
Process tree

Container
└── [1]  piggie
```

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 
 	metadata "github.com/checkpoint-restore/checkpointctl/lib"
 	"github.com/spf13/cobra"
@@ -17,6 +18,7 @@ var (
 	stats     bool
 	mounts    bool
 	fullPaths bool
+	psTree    bool
 	showAll   bool
 )
 
@@ -63,6 +65,12 @@ func setupShow() *cobra.Command {
 		"mounts",
 		false,
 		"Print overview about mounts used in the checkpoints",
+	)
+	flags.BoolVar(
+		&psTree,
+		"ps-tree",
+		false,
+		"Print overview about the process tree in the checkpoints",
 	)
 	flags.BoolVar(
 		&fullPaths,
@@ -122,10 +130,21 @@ func show(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
+	// A list of files that need to be unarchived. The files need not be
+	// full paths. Even a substring of the file name is valid.
 	files := []string{metadata.SpecDumpFile, metadata.ConfigDumpFile}
 
 	if stats {
 		files = append(files, "stats-dump")
+	}
+
+	if psTree {
+		files = append(
+			files,
+			filepath.Join(metadata.CheckpointDirectory, "pstree.img"),
+			// All core-*.img files
+			filepath.Join(metadata.CheckpointDirectory, "core-"),
+		)
 	}
 
 	if err := untarFiles(input, dir, files); err != nil {
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -122,7 +122,13 @@ func show(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
-	if err := untarFiles(input, dir, metadata.SpecDumpFile, metadata.ConfigDumpFile); err != nil {
+	files := []string{metadata.SpecDumpFile, metadata.ConfigDumpFile}
+
+	if stats {
+		files = append(files, "stats-dump")
+	}
+
+	if err := untarFiles(input, dir, files); err != nil {
 		return err
 	}
 
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -96,6 +96,7 @@ func show(cmd *cobra.Command, args []string) error {
 	if showAll {
 		stats = true
 		mounts = true
+		psTree = true
 	}
 	if fullPaths && !mounts {
 		return fmt.Errorf("Cannot use --full-paths without --mounts/--all option")
--- a/container.go
+++ b/container.go
@@ -234,16 +234,17 @@ type archiveSizes struct {
 func getArchiveSizes(archiveInput string) (*archiveSizes, error) {
 	result := &archiveSizes{}
 
-	err := iterateTarArchive(archiveInput, func(hdr *tar.Header) {
-		if hdr.FileInfo().Mode().IsRegular() {
-			if hasPrefix(hdr.Name, metadata.CheckpointDirectory) {
+	err := iterateTarArchive(archiveInput, func(r *tar.Reader, header *tar.Header) error {
+		if header.FileInfo().Mode().IsRegular() {
+			if hasPrefix(header.Name, metadata.CheckpointDirectory) {
 				// Add the file size to the total checkpoint size
-				result.checkpointSize += hdr.Size
-			} else if hasPrefix(hdr.Name, metadata.RootFsDiffTar) {
+				result.checkpointSize += header.Size
+			} else if hasPrefix(header.Name, metadata.RootFsDiffTar) {
 				// Read the size of rootfs diff
-				result.rootFsDiffTarSize = hdr.Size
+				result.rootFsDiffTarSize = header.Size
 			}
 		}
+		return nil
 	})
 	return result, err
 }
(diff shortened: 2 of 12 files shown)


go.mod
```diff
@@ -13,16 +13,17 @@ require (
 require (
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
-	github.com/klauspost/compress v1.16.4 // indirect
-	github.com/klauspost/pgzip v1.2.5 // indirect
-	github.com/mattn/go-runewidth v0.0.9 // indirect
+	github.com/klauspost/compress v1.16.5 // indirect
```
rst0git, line 16: @snprajwal Could you move this patch in a separate pull request?

snprajwal:
CRIU dump and restore are failing in the CI, I'm not sure why. Is it because of insufficient permissions or something? I based the workflow directly off of the one we have in go-criu.

adrianreber:
> CRIU dump and restore are failing in the CI, I'm not sure why. Is it because of insufficient permissions or something? I based the workflow directly off of the one we have in go-criu.

We run in a unprivileged container. GitHub actions supports privileged containers. Please try that.

snprajwal:
> We run in a unprivileged container. GitHub actions supports privileged containers. Please try that.

Looks like that helped - now both workflows fail on restore :laughing:

adrianreber:
> > We run in a unprivileged container. GitHub actions supports privileged containers. Please try that.
> 
> Looks like that helped - now both workflows fail on restore :laughing: 

You either need to print the logs in an error case or log to stdout.

It could be necessary to mount /lib/modules from the host into the container so that necessary module can be loaded.

But without logs it is difficult to tell.

snprajwal:
I don't want to create too much noise on this PR. Let me try it out on a separate branch, and push the changes here once I get it working

adrianreber:
I think we might need to wait a bit with this PR. My favourite approach right now would be to make everything PID tree based. Maybe the table view is only useful for the quick overview. I am curious in how far the tree based approach would work for `checkpointctl` overall.

adrianreber:
We talked about binary sizes before. I think we should include a CI check about the binary size like Podman has. Build each PR without the PR and record the binary size, build the PR with the changes and record the binary size. Compare both sizes and fail if +10%.

snprajwal:
I've added the process tree logic here. I didn't want to create a separate branch and PR for it as of now since I had important changes in the first two commits, and did not want to duplicate them. If this looks good, we can discuss how file descriptors, memory maps, etc. integrate into this and add them accordingly. I still have the commit and code for the file descriptors feature saved locally.

adrianreber:
There are a couple of things that should be done differently. This PR mixes too many things. There is some renaming going on which should be a separate commit. The changes to the testing and the new features should be different PRs.

For the new feature I would like to see some example output in the PR description.

Instead of the loop.c example, let's try to use `piggie.c` from go-criu, which also creates a couple of namespaces as this will be helpful at some point during testing.

Also the binary size comparison PR should be done before any more includes from go-criu.

snprajwal:
Got it, I'll split it up and open separate PRs.

checkpointctl.go
```diff
@@ -47,22 +48,28 @@ func setupShow() *cobra.Command {
 	}
 	flags := cmd.Flags()
 	flags.BoolVar(
-		&printStats,
+		&stats,
 		"print-stats",
 		false,
-		"Print checkpointing statistics if available",
+		"Print checkpointstatistics if available",
```
rst0git, line 54: "checkpoint statistics" are two words

snprajwal: Ack I think I messed up while rebasing. I'll fix it when I rebase after #61 is merged.

snprajwal:
@adrianreber I've updated the PR description with the tree we currently display during tests (with piggie).

adrianreber:
Looks good for a first step. I like it.

container.go
```diff
@@ -117,12 +118,23 @@ func showContainerCheckpoint(checkpointDirectory, input string) error {
...
 		if err != nil {
-			return fmt.Errorf("unable to display checkpointing statistics: %w", err)
+			return fmt.Errorf("failed to get dump statistics: %w", err)
 		}
 
 		renderDumpStats(dumpStats)
 	}
 
+	if psTree {
+		c := crit.New("", "", checkpointDirectory, false, false)
```
rst0git, line 128: > Error: failed to get process tree: Error opening binary file: open /tmp/checkpointctl1506966507/pstree.img: no such file or directory

snprajwal: Ack, figured this out. The temporary directory being created is one level above the checkpoint directory. I need to append that path here.

snprajwal: I tried listing the files present in the temporary directory, and got this:
```
/tmp/checkpointctl65800022/config.dump
/tmp/checkpointctl65800022/spec.dump
/tmp/checkpointctl65800022/stats-dump
```

checkpointctl.go
```diff
@@ -122,7 +130,9 @@ func show(cmd *cobra.Command, args []string) error {
 		}
 	}()
 
-	if err := untarFiles(input, dir, metadata.SpecDumpFile, metadata.ConfigDumpFile); err != nil {
+	// List of files required from the checkpoint archive
+	files := []string{metadata.SpecDumpFile, metadata.ConfigDumpFile, metadata.CheckpointDirectory}
```
adrianreber, line 134: This works, but this is against the goal I had in mind with the change to only unpack the needed files.

If the user requests the process tree, only then should the corresponding files be unpacked. According to the test case below only `pstree.img` and `core-*.img` files are needed.

container.go
```diff
@@ -257,38 +295,15 @@ func shortenPath(path string) string {
...
-
-	// Remove files from ExcludePatterns if they are present
-	for _, file := range files {
-		for i, pattern := range options.ExcludePatterns {
-			if pattern == file {
-				options.ExcludePatterns = append(options.ExcludePatterns[:i], options.ExcludePatterns[i+1:]...)
-				break
-			}
-		}
+		IncludeFiles: files,
```
adrianreber, line 306: I understand what you are trying to do, but I am not sure it actually works as expected. Did you try it out?

When I last used this library, `IncludeFiles` did not work as I expected it during unpacking.

snprajwal: Last I checked, it worked as expected. Let me double check once more.

checkpointctl.go
snprajwal: Got it, I'll change this. But do note that we will have to add more files as we go along, for the other explore types.

snprajwal: I have changed it to include the individual files that each feature requires. For now, we are only unarchiving `pstree.img` and `core-*.img` for the process tree. A nifty feature of `IncludeFiles` with `tar` is that it allows us to specify wildcards!

test/checkpointctl.bats
```diff
@@ -113,7 +113,7 @@ function teardown() {
 	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/test.tar . )
 	checkpointctl show "$TEST_TMP_DIR2"/test.tar --stats
 	[ "$status" -eq 1 ]
-	[[ ${lines[6]} == *"unable to display checkpointing statistics"* ]]
+	[[ ${lines[6]} == *"failed to get dump statistics"* ]]
```
rst0git, line 116: Is this change intentional? (It looks unrelated)

container.go
```diff
@@ -221,6 +228,31 @@ func renderDumpStats(dumpStats *images.DumpStatsEntry) {
 	table.Render()
 }
 
+func renderPsTree(psTree *crit.PsTree, containerName string) {
+	var tree treeprint.Tree
+	if containerName != "" {
+		tree = treeprint.NewWithRoot(containerName)
+	} else {
+		tree = treeprint.NewWithRoot("Container")
+	}
```
rst0git, line 237: Could we simplify this condition to something like the following?
```Go
if containerName == "" {
	containerName = "Container"
}
tree = treeprint.NewWithRoot(containerName)
```

rst0git (approved):

container.go
snprajwal: Thanks for pointing this out. I realised now that during tests, we move only the required files into the checkpoint folder before creating the archive. Since I used that to test this functionality, it appeared as though only the necessary files are being unpacked. I'll fix this.

adrianreber (approved):

PR #57, opened 2023-05-28
chore: update dependencies

snprajwal:
Just bumps up a bunch of deps to the newer versions.


adrianreber:
How did you change those version dependencies?

Don't we have dependabot to do these changes.

Should we really change indirect dependencies?

adrianreber:
Do we have something like `make vendor` to verify that the dependencies are correct during CI?

snprajwal:
I updated the dependencies by running `go get -u`, and then vendored it with `go mod tidy` + `go mod vendor`. I don't think Dependabot tracks indirect dependencies.

snprajwal:
> Do we have something like `make vendor` to verify that the dependencies are correct during CI?

Doesn't look like it, let me add it. Would you want it in this PR itself, or a different PR?

adrianreber:
> I updated the dependencies by running `go get -u`, and then vendored it with `go mod tidy` + `go mod vendor`. I don't think Dependabot tracks indirect dependencies.

Should we really change indirect dependencies? I am not yet convinced. We have those because other projects pull them in and I am not sure it is correct to update those if the vendored in projects use something else.

adrianreber:
> > Do we have something like `make vendor` to verify that the dependencies are correct during CI?
> 
> Doesn't look like it, let me add it. Would you want it in this PR itself, or a different PR?

Separate PR sounds good.

Something like go mod tidy vendor verify and then maybe a git diff to verify nothing changed.

Please have a look at other go projects how they verify stable dependencies in CI.

snprajwal:
> We have those because other projects pull them in and I am not sure it is correct to update those if the vendored in projects use something else.

AFAIK, doing `go get -u` updates the indirect dependencies to the version that the direct dependencies themselves use. So they are changed only if one of our direct deps has bumped them up in the respective `go.mod`.

adrianreber:
I am still skeptical about updating indirect dependencies. That sounds unnecessary.

rst0git:
> > We have those because other projects pull them in and I am not sure it is correct to update those if the vendored in projects use something else.
> 
> AFAIK, doing `go get -u` updates the indirect dependencies to the version that the direct dependencies themselves use. So they are changed only if one of our direct deps has bumped them up in the respective `go.mod`.

From the documentation it looks like `go get -u`  attempts to upgrade all direct and indirect (transitive) dependencies[^1], and indirect dependencies can be updated beyond what is required by the direct dependencies.

[^1]: https://github.com/golang/go/wiki/Modules

adrianreber:
@rst0git so, are in favour of upgrading indirect dependencies or not? :wink:

rst0git:
> so, are in favour of upgrading indirect dependencies or not?

I also think it may be unnecessary to upgrade the indirect dependencies unless we have a specific reason to do so.

PR #58, opened 2023-05-28
ci: verify go.mod and vendor directory

snprajwal:
This PR also consolidates the linters into a single "verify" workflow, along the same lines as go-criu.


.github/workflows/verify.yml
```diff
@@ -13,6 +13,15 @@ jobs:
           version: latest
           only-new-issues: true
 
+  gomod:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: verify go.mod/go.sum
+      run: |
+        go mod tidy
```
rst0git, line 22: Should we run `make vendor` instead?

adrianreber: Sounds like a good idea.

rst0git:
Should we configure dependbot with [`dependency-type: "all"`](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#allow)?

```yaml
    allow:
      # Allow both direct and indirect updates for all packages
      - dependency-type: "all"
```

.github/workflows/verify.yml
```diff
@@ -0,0 +1,33 @@
...
+    - uses: actions/checkout@v3
+    - uses: dorny/paths-filter@v2
+      id: changes
+      with:
+        filters: |
+          md:
+            - 'README.md'
+    - name: Lint markdown
+      if: steps.changes.outputs.md == 'true'
+      uses: DavidAnson/markdownlint-cli2-action@v9
```
rst0git, line 39: Should we also update to `DavidAnson/markdownlint-cli2-action@v10`?

.github/workflows/verify.yml
```diff
@@ -0,0 +1,33 @@
...
+    permissions:
+      pull-requests: read
+    steps:
+    - uses: actions/checkout@v3
+    - uses: dorny/paths-filter@v2
+      id: changes
+      with:
+        filters: |
+          md:
+            - 'README.md'
```
rst0git, line 36: It might make sense to run this when `.github/workflows/verify.yml` has been modified as well.

rst0git (approved):
LGTM

.github/workflows/verify.yml
adrianreber: Why? I don't get it.

snprajwal: If we decide to change the workflow, or bump up the version of markdownlint-cli2, we would want to run the check, even if the markdown files in the repo haven't changed.

PR #59, opened 2023-06-04
ci: add job to check increase in binary size

snprajwal:
This job checks the difference in binary size across commits in a PR, and fails if the difference is more than a set maximum limit. This allows us to detect early on if there is any possible bloating in the binary due to poorly structured dependencies. The job can be overridden by adding a `bloat-ok` label to the PR.

### DIFF
--- a/.github/workflows/check-size.yml
+++ b/.github/workflows/check-size.yml
@@ -1,0 +1,22 @@
+name: Check binary size
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled, ready_for_review]
+
+jobs:
+  check_size:
+    runs-on: ubuntu-latest
+    # Check if pull request does not have label "bloat-ok".
+    if: "!contains(github.event.pull_request.labels.*.name, 'bloat-ok')"
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        # Use HEAD from the PR instead of the merge commit
+        ref: ${{ github.event.pull_request.head.sha }}
+        # Needed to rebase against the base branch
+        fetch-depth: 0
+    - name: Configure base branch without switching current branch
+      run: git fetch origin ${GITHUB_BASE_REF}:${GITHUB_BASE_REF}
+    - name: Compare binary size change across each commit
+      run: git rebase ${GITHUB_BASE_REF}^ -x test/check-size.sh
--- a/test/check-size.sh
+++ b/test/check-size.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# A shell script to check if the binary size
+# has increased beyond a certain threshold
+# across commits. Meant to be used in CI with
+# git rebase <base branch>^ -x check-size.sh
+
+# Fail fast on errors
+set -e
+
+BIN_NAME=checkpointctl
+PREV_SIZE_FILE=prev_size
+# Maximum allowable size difference, in bytes
+MAX_DIFF=51200
+
+# Build the checkpointctl binary. If the commit is not self-contained,
+# the build will fail, in which case there is no point checking for a
+# change in the size of the binary.
+make
+# Store the binary size
+BIN_SIZE=$(stat -c%s "$BIN_NAME")
+
+if [[ -f "$PREV_SIZE_FILE" ]]; then
+	# Read the previous size from the file
+	PREV_SIZE=$(cat "$PREV_SIZE_FILE")
+	# Calculate the difference between current and previous size
+	DIFF=$((BIN_SIZE - PREV_SIZE))
+	if [[ $DIFF -gt $MAX_DIFF ]]; then
+		echo "FAIL: size difference of \"$DIFF\" B exceeds limit \"$MAX_DIFF\" B"
+		exit 1
+	else
+		echo "PASS: size difference of \"$DIFF\" B within limit \"$MAX_DIFF\" B"
+	fi
+else
+	# This means this is the first run of the script.
+	# The original file size will be stored, and used
+	# to compare against subsequent values ahead.
+	echo "No previous size present, storing current size"
+	echo "$BIN_SIZE" > "$PREV_SIZE_FILE"
+fi


snprajwal:
`fatal: invalid upstream 'main^'`

Slightly weird. This works locally, even with the single quotes. I'll debug it on a separate branch.

snprajwal:
Right, now the workflow can't find the `test/check-size.sh` file.

adrianreber:
Looks good. Can you make CI happy using the `bloat-ok` label?

snprajwal:
I don't seem to have the rights to add labels to the PR

adrianreber:
I added the label. It still runs the job. I thought it should ignore it now.

snprajwal:
I thought the same too 🤔 weird behaviour

.github/workflows/tests.yml
```diff
@@ -31,6 +31,23 @@ jobs:
       - name: Run make install/uninstall
         run: test/uninstall.sh
 
+  check_size:
+    if: |
+      github.event_name == 'pull_request' &&
+      !contains(github.event.pull_request.labels.*.name, 'bloat-ok')
```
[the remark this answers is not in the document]
snprajwal: Makes sense, I'll move it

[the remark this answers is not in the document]
snprajwal: > If you want to run your workflow when a label is added to or removed from an issue, pull request, or discussion, use the labeled or unlabeled activity types for the [issues](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#issues), [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request), [pull_request_target](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target), or [discussion](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#discussion) events instead.

I think this is what we need

.github/workflows/tests.yml
```diff
@@ -31,6 +31,23 @@ jobs:
...
+    - uses: actions/checkout@v3
+      with:
+        # Use HEAD from the PR instead of the merge commit
+        ref: ${{ github.event.pull_request.head.sha }}
+        # Needed to rebase against the base branch
+        fetch-depth: 0
+    - name: Configure base branch without switching current branch
+      run: git fetch origin ${GITHUB_BASE_REF}:${GITHUB_BASE_REF}
+    - name: Compare binary size change across each commit
+      run: git rebase ${GITHUB_BASE_REF}^ -x ../../test/check-size.sh
```
rst0git, line 49: > fatal: cannot run ../../test/check-size.sh: No such file or directory

test/check-size.sh
```diff
@@ -0,0 +1,34 @@
...
+	# Calculate the difference between current and previous size
+	DIFF=$((BIN_SIZE - PREV_SIZE))
+	if [[ $DIFF -gt $MAX_DIFF ]]; then
+		echo "FAIL: size difference of \"$DIFF\" B exceeds limit \"$MAX_DIFF\" B"
+		exit 1
+	else
+		echo "PASS: size difference of \"$DIFF\" B within limit \"$MAX_DIFF\" B"
+	fi
+else
+	echo "No previous size present, storing current size"
```
rst0git, line 30: Here it might be good to add a comment like "_First time through: preserve original file size_",
similar to the [make-and-check-size](https://github.com/containers/podman/blob/aee7a3c16aee82d11689db91e42f82e52a690d98/hack/make-and-check-size#LL132C9-L132C58) script.

.github/workflows/tests.yml
snprajwal: Yes, I mentioned that in the comment above. I tried a few paths, including using the workspace location with `$GITHUB_WORKSPACE/test/check-size.sh`, but the file is not being found for some reason.

adrianreber: I think the error makes sense because you are checking out a version which does not have the script yet.

We should be able to ignore this for the initial commit using the mentioned label.

rst0git: Have you tried just `test/check-size.sh`? If this path is not found it is most likely because `git rebase ${GITHUB_BASE_REF}` would checkout to the "main" branch. Since the pull request is not yet merged, the "main" branch does not contain this file.

snprajwal:
Nice, the workflow doesn't run since the PR is labeled. Can either of you please remove the label so we can see if it gets triggered. Also, I've added a line to test if the script is found by the CI shell. If that works, I'll remove and force push so that we can merge this

snprajwal:
Great! The script is present, getting rid of the test and pushing.

snprajwal:
:thinking: why did it not run now

snprajwal:
Ah, got it. I overrode the default triggers for PR events.

snprajwal:
Now we're good to go! :tada:

.github/workflows/check-size.yml
```diff
@@ -0,0 +1,22 @@
+name: Check binary size
+
+on:
+  pull_request:
+    types: [opened, reopened, synchronize, labeled, unlabeled]
```
rst0git, line 5: Should we also add `ready_for_review`?

snprajwal: I don't think that's ideal, because we would want this to run on commits on draft PRs too. It would help identify bloating early on and change the direction of development accordingly. As far as I understand, `ready_for_review` is triggered when a draft PR is converted into a regular one. By then, it might be too late to notify this to the author.

.github/workflows/check-size.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: actions/checkout@v3
+      with:
+        # Use HEAD from the PR instead of the merge commit
+        ref: ${{ github.event.pull_request.head.sha }}
+        # Needed to rebase against the base branch
+        fetch-depth: 0
+    - name: Configure base branch without switching current branch
+      run: git fetch origin ${GITHUB_BASE_REF}:${GITHUB_BASE_REF}
+    - name: Compare binary size change across each commit
+      run: git rebase ${GITHUB_BASE_REF}^ -x test/check-size.sh
```
rst0git, line 22: Do we need to run `check-size.sh` here twice -- with and without `git rebase ${GITHUB_BASE_REF}^`?

snprajwal: I'm not sure I understand. The `rebase -x` lets us detect which commit is the offender, whereas running against HEAD without that will only give us a yes/no for the bloating.

rst0git: The first run of `check-size.sh` it would create `prev_size` with the size of the binary from the _main_ branch.
The second run would build a binary with commits from the pull request and compare the size.

.github/workflows/check-size.yml
rst0git: `ready_for_review` should not prevent this workflow from running on draft PRs.

.github/workflows/check-size.yml
snprajwal: Ah, that would require a checkout to main, and back to the PR. Instead, doing `rebase main^` with the `^` lets us use the parent commit from main as the first one in the rebase process.

.github/workflows/check-size.yml
snprajwal: Ah understood. But that wouldn't change the contents of the PR itself, so wouldn't it be a redundant run? If the workflow has already run on the latest commit before converting from a draft PR to a regular one, then we'll be doing a duplicate run.

.github/workflows/check-size.yml
rst0git: For example, running the following commit executes only the `else` statement within the script:
```
$ git rebase main^ -x test/check-size.sh
Executing: test/check-size.sh
go build -buildmode=pie -o checkpointctl -ldflags "-X main.name=checkpointctl -X main.version=0.1.0"
No previous size present, storing current size
Successfully rebased and updated refs/heads/check-bin-size.
```

.github/workflows/check-size.yml
rst0git: Running the workflow when converting from a draft PR might be helpful because it would pick up any changes to the main branch since the last run.

.github/workflows/check-size.yml
snprajwal: :thinking: I'm not sure how you managed to make it run just once. When I tried running `git rebase main^ -x test/check-size.sh` on a branch one commit ahead of `check-bin-size`, the script ran twice - once with just the `else` part, and a second time for the new commit to compare the size. I borrowed the command itself straight from podman, so I'm quite sure it covers all commits, including the latest one. Still not understanding why we would have to run the script by itself once again :/

rst0git: Ah, you are right. I ran the command on the `check-bin-size` branch while having the same commits applied to the main branch. Since the head commits were the same `git rebase` executed the script only once.

I was tested again on another branch with different commits and it works as expected. Thanks for pointing this out.

rst0git (approved):

PR #60, opened 2023-06-04
refactor: move table rendering into own funcs

snprajwal:
The logic for rendering tables with the checkpoint data was all present
in a single function `showContainerCheckpoint()`. This has now been
refactored to provide each table its own function that can be called to
render it. This increases code maintainability going forward as more and
more tables are added to the `checkpointctl show` command.

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -12,12 +12,12 @@ import (
 )
 
 var (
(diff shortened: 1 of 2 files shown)


rst0git (approved):
LGTM

PR #62, opened 2023-06-08
fix(ci): set git credentials for rebase

snprajwal:
The workflow fails during rebase since git cannot find any details of a user to create the commits. This adds a dummy user to fix this.

### DIFF
--- a/.github/workflows/check-size.yml
+++ b/.github/workflows/check-size.yml
@@ -12,8 +12,6 @@ jobs:
     steps:
     - uses: actions/checkout@v3
       with:
-        # Use HEAD from the PR instead of the merge commit
-        ref: ${{ github.event.pull_request.head.sha }}
         # Needed to rebase against the base branch
         fetch-depth: 0
     - name: Configure git user details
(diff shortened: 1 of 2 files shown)


snprajwal:
I discovered this only after I rebased the other PRs :/

PTAL @adrianreber

adrianreber:
Thanks for the quick fix. Do you understand the output:
```
Rebasing (2/4)

                                                                                
Executing: test/check-size.sh
go build -buildmode=pie -o checkpointctl -ldflags "-X main.name=checkpointctl -X main.version=0.1.0"
No previous size present, storing current size
Rebasing (3/4)
Rebasing (4/4)

                                                                                
Executing: test/check-size.sh
make: Nothing to be done for 'all'.
PASS: size difference of "0" B within limit "51200" B

                                                                                
Successfully rebased and updated detached HEAD.
```

Why does rebase say `/4`. Shouldn't be just `/2`?

Maybe doing `git rebase -v` would be helpful to see the commit IDs.

Also, can you update the message from the script to also print the actual size of the binary. Not just the difference but the current value for each run.

adrianreber:
Also, maybe, instead of `prev_size` maybe:
```
PREV_SIZE_FILE=$(mktemp)
trap 'rm -f ${PREV_SIZE_FILE}' EXIT QUIT HUP TERM
```
Not 100% sure about the syntax.

adrianreber:
> Also, maybe, instead of `prev_size` maybe:
> 
> ```
> PREV_SIZE_FILE=$(mktemp)
> trap 'rm -f ${PREV_SIZE_FILE}' EXIT QUIT HUP TERM
> ```
> 
> Not 100% sure about the syntax.

Ah, that doesn't work. Forget that.

snprajwal:
Did you figure out why the number of commits being rebased is different? I would think it's because of the merge commits from main

PR #63, opened 2023-06-08
test/check-size: show error if make fails

rst0git:
This pull request adds an error message when compilation fails when running the `check-size.sh` script.

### DIFF
--- a/.github/workflows/check-size.yml
+++ b/.github/workflows/check-size.yml
@@ -14,6 +14,8 @@ jobs:
       with:
         # Needed to rebase against the base branch
         fetch-depth: 0
+        # Checkout pull request HEAD commit instead of merge commit
+        ref: ${{ github.event.pull_request.head.sha }}
     - name: Configure git user details
       run: |
         git config --global user.email "checkpoint-restore@users.noreply.github.com"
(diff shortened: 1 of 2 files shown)


.github/workflows/check-size.yml
```diff
@@ -16,8 +16,8 @@ jobs:
         fetch-depth: 0
```
snprajwal, line 16: Can we go back to using the PR HEAD instead of the merge commit? It'll keep the commit SHA consistent with the branch in the error messages from the script.
```suggestion
        fetch-depth: 0
        # Use the HEAD commit from the PR instead of merge commit
        ref: ${{ github.event.pull_request.head.sha }}
```

rst0git: I'm not sure if I understand this. According to the [docs](https://github.com/marketplace/actions/checkout#usage), `ref` specifies the branch, tag or SHA to checkout. The default is the reference or SHA for that event that triggered a workflow.

snprajwal: By default, the SHA that is checked out is a merge commit from the base branch on the PR. To use the PR HEAD instead of this merge commit, we need to explicitly specify the SHA from the event context provided by GitHub. Ref: https://github.com/marketplace/actions/checkout#Checkout-pull-request-HEAD-commit-instead-of-merge-commit

snprajwal (approved):

PR #64, opened 2023-06-08
ci: use Ubuntu for tests instead of Fedora

snprajwal:
Previously, the Fedora container image was being used to test checkpointctl since the Ubuntu image had an outdated version of bats. This does not affect the tests themselves, as the coverage workflow continues to use `ubuntu-latest` without issues. This change will eliminate the need for having a privileged container to run criu while generating image files for tests.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,12 +5,14 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    container:
-      image: registry.fedoraproject.org/fedora:latest
     steps:
       - uses: actions/checkout@v3
       - name: Install tools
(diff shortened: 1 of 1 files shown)


adrianreber:
I prefer distribution packages instead of installing from a third source.

snprajwal:
I agree, but the `bats` package on Ubuntu has been updated for 22.10 and 23.04. Unfortunately, the GitHub image uses 22.04. I think using the NPM package is a less expensive workaround to running a container with a different distribution.

adrianreber:
In addition to bats being too old in Ubuntu, the same is true for shellcheck and CRIU doesn't work at all. So we need additional repositories for every package and Fedora just has the correct version of all of it.

snprajwal:
Ah understood. This is unnecessary then, let's close it.

PR #65, opened 2023-06-10
Build static checkpointctl binary for improved portability

rst0git:
This pull request modifies the Makefile to build a static `checkpointctl` binary and eliminate dependency on dynamic libraries. This change ensures that the resulting binary can be effortlessly distributed and executed across various systems and linux distributions. For example, a binary built on a Fedora system will seamlessly run on Ubuntu and vice versa.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,11 +94,15 @@ help:
 	@echo "Usage: make <target>"
 	@echo " * clean - remove artifacts"
 	@echo " * lint - verify the source code (shellcheck/golangci-lint)"
-	@echo " * golang-lint - run golang-lint"
-	@echo " * shellcheck - run shellecheck"
-	@echo " * vendor - update go.mod, go.sum and vendor directory"
+	@echo " * golang-lint - run golangci-lint"
+	@echo " * shellcheck - run shellcheck"
+	@echo " * vendor - update go.mod, go.sum, and vendor directory"
 	@echo " * test - run tests"
 	@echo " * test-junit - run tests and create junit output"
+	@echo " * coverage - generate test coverage report"
+	@echo " * codecov - upload coverage report to codecov.io"
+	@echo " * install - install the binary to $(BINDIR)"
+	@echo " * uninstall - remove the installed binary from $(BINDIR)"
 	@echo " * help - show help"
 
 .PHONY: clean install uninstall lint golang-lint shellcheck vendor test help check-go-version test-junit
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ $(NAME).coverage: check-go-version $(GO_SRC)
 		-buildmode=pie -o $@ \
 		-ldflags "-X main.name=$(NAME) -X main.version=${VERSION}"
 
+release:
+	CGO_ENABLED=0 $(GO_BUILD) -o $(NAME) -ldflags "-X main.name=$(NAME) -X main.version=${VERSION}"
 
 install: $(NAME)
 	@echo "  INSTALL " $<
@@ -103,6 +105,7 @@ help:
 	@echo " * codecov - upload coverage report to codecov.io"
 	@echo " * install - install the binary to $(BINDIR)"
 	@echo " * uninstall - remove the installed binary from $(BINDIR)"
+	@echo " * release - build a static binary"
 	@echo " * help - show help"
 
-.PHONY: clean install uninstall lint golang-lint shellcheck vendor test help check-go-version test-junit
+.PHONY: clean install uninstall release lint golang-lint shellcheck vendor test help check-go-version test-junit
--- a/Makefile
+++ b/Makefile
@@ -37,12 +37,12 @@ check-go-version:
 
 
 $(NAME): $(GO_SRC)
-	$(GO_BUILD) -buildmode=pie -o $@ -ldflags "-X main.name=$(NAME) -X main.version=${VERSION}"
+	$(GO_BUILD) -o $@ -ldflags "-X main.name=$(NAME) -X main.version=${VERSION}"
 
 $(NAME).coverage: check-go-version $(GO_SRC)
 	$(GO) build \
 		-cover \
-		-buildmode=pie -o $@ \
+		-o $@ \
 		-ldflags "-X main.name=$(NAME) -X main.version=${VERSION}"
 
 release:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -43,3 +43,23 @@ jobs:
         name: event-file
         retention-days: 1
         path: ${{ github.event_path }}
+
+  release:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v3
+
+    - name: Build release binary
+      run: make release
+
+    - name: Check static binary
+      run: |
+        file_output=$(file checkpointctl)
+        if [[ "$file_output" =~ .*statically\ linked.* ]]; then
+          echo "Binary is static"
+          exit 0
+        else
+          echo "Binary is not static"
+          exit 1
+        fi


adrianreber:
Hmm...
```
 CGO_ENABLED=0 go build -a -o checkpointctl -ldflags "-X main.name=checkpointctl -X main.version=0.1.0"
BINARY SIZE (e4432ed): 10457771
PASS: size difference of -610717 B within limit 51200 B
Successfully rebased and updated detached HEAD.
```
Your binary gets smaller. This is not what I would expect from a static binary...

Overall I am sceptical regarding this change. If we want to distribute binaries for a release for easy download, then sure, let's create a static binary, but for local use I do not think we need static binaries. I would have expected that the static binary is bigger, which it does not seem to be. Which is confusing. I also read somewhere that compilation with `CGO_ENABLED=0` is slower. Did you compare compile times?

snprajwal:
We can (somewhat) statically compile checkpointctl with the following command:

`go build -buildmode=pie -o checkpointctl -ldflags "-X main.name=checkpointctl -X main.version=0.1.0 -extldflags=-static"`

`-extldflags` is passed to the external linker (`ld`), which will statically link the C libraries. But building with this command gives us warnings:

```
/usr/bin/ld: /tmp/go-link-161501320/000002.o: in function `mygetgrouplist':
/_/os/user/getgrouplist_unix.go:15: warning: Using 'getgrouplist' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-161501320/000001.o: in function `mygetgrgid_r':
/_/os/user/cgo_lookup_cgo.go:45: warning: Using 'getgrgid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-161501320/000001.o: in function `mygetgrnam_r':
/_/os/user/cgo_lookup_cgo.go:54: warning: Using 'getgrnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-161501320/000001.o: in function `mygetpwnam_r':
/_/os/user/cgo_lookup_cgo.go:36: warning: Using 'getpwnam_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-161501320/000001.o: in function `mygetpwuid_r':
/_/os/user/cgo_lookup_cgo.go:27: warning: Using 'getpwuid_r' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
/usr/bin/ld: /tmp/go-link-161501320/000011.o: in function `_cgo_cbcce81e6342_C2func_getaddrinfo':
/tmp/go-build/cgo_unix_cgo.cgo2.c:58: warning: Using 'getaddrinfo' in statically linked applications requires at runtime the shared libraries from the glibc version used for linking
```

This could possibly be due to glibc using libnss for address resolution, and libnss cannot be statically linked (see [here](https://stackoverflow.com/a/3087067 for details).

Also, I never bothered to question it till now, but why are we building checkpointctl as a PIE?

adrianreber:
> Also, I never bothered to question it till now, but why are we building checkpointctl as a PIE?

I probably copied it from some other Go project. Do you see it as a problem?

snprajwal:
Not particularly, but it seems unnecessary. Wouldn't we be adding extra overhead in the binary to handle relocation?

rst0git:
>  If we want to distribute binaries for a release for easy download, then sure, let's create a static binary, but for local use I do not think we need static binaries.

@adrianreber Would it make more sense to create a new makefile target (e.g., "release" or "static") that builds a static binary?

adrianreber:
Sounds good

rst0git:
> Wouldn't we be adding extra overhead in the binary to handle relocation?

Indeed, it looks like removing the `-buildmode=pie` is reducing the size of the binary from 11M to 9.8M.

snprajwal:
@rst0git would you want to revert 71f18e4 since using `CGO_ENABLED=0` is unviable for us?

rst0git:
I've updated the pull request to replace `CGO_ENABLED=0` with the corresponding build flags for static linking.

avagin:
>> I also read somewhere that compilation with CGO_ENABLED=0 is slower.

> Correct, the Go standard library is built without any flags. If CGo is disabled, it needs to be rebuilt.

Why do we care about the compilation time? It isn't dramatically slower. CGO_ENABLED=0 should be preferred rather than linking with static glibc and others...

snprajwal:
> > > I also read somewhere that compilation with CGO_ENABLED=0 is slower.
> 
> > Correct, the Go standard library is built without any flags. If CGo is disabled, it needs to be rebuilt.
> 
> Why do we care about the compilation time? It isn't dramatically slower. CGO_ENABLED=0 should be preferred rather than linking with static glibc and others...

The compilation time was more of a side point, the main reason we can't use `CGO_ENABLED=0` is because it doesn't compile the Go files that call C code.

Makefile
```diff
@@ -94,11 +97,16 @@ help:
...
-	@echo " * shellcheck - run shellecheck"
-	@echo " * vendor - update go.mod, go.sum and vendor directory"
+	@echo " * golang-lint - run golangci-lint"
+	@echo " * shellcheck - run shellcheck"
+	@echo " * vendor - update go.mod, go.sum, and vendor directory"
 	@echo " * test - run tests"
 	@echo " * test-junit - run tests and create junit output"
+	@echo " * coverage - generate test coverage report"
+	@echo " * codecov - upload coverage report to codecov.io"
+	@echo " * install - install the binary to $(DESTDIR)$(BINDIR)"
```
snprajwal, line 107:
```
PREFIX ?= $(DESTDIR)/usr/local
BINDIR ?= $(PREFIX)/bin
```

If we're doing this, why do we need `$(DESTDIR)$(BINDIR)` again?

rst0git:
> > Why do we care about the compilation time? It isn't dramatically slower. CGO_ENABLED=0 should be preferred rather than linking with static glibc and others...
> 
> The compilation time was more of a side point, the main reason we can't use `CGO_ENABLED=0` is because it doesn't compile the Go files that call C code.

I don't think there is a big difference between using `CGO_ENABLED=0` or `-tags netgo,osusergo -ldflags "-linkmode 'external' -extldflags '-static'"` but from what I have [read](https://mt165.co.uk/blog/static-link-go/) it seems that `CGO_ENABLED=0` is the preferred way of building a static binary. By default in [Go version 1.20](https://go.dev/doc/go1.20#cgo), `CGO_ENABLED` would be set to 0 when the `CC` environment variable is unset and the default C compiler is not found in the path.

avagin:
> The compilation time was more of a side point, the main reason we can't use `CGO_ENABLED=0` is because it doesn't compile the Go files that call C code.

Do you have any C code in this project?

snprajwal:
> Do you have any C code in this project?

No, but some of the deps do (x/sys, containers/storage)

avagin:
@snprajwal I know nothing about containers/storage, but x/sys works fine without cgo, we use it in gVisor and runsc compiled without CGO.

avagin:
Here is an example of how containers/storage implements no-cgo alternatives:
https://github.com/containers/storage/blob/c989f8f2d4ee28bf3d181beffaa1c09341545758/pkg/unshare/getenv_linux_nocgo.go#L9

snprajwal:
> At the moment for checkpointctl, I don't think there is a big difference between using `CGO_ENABLED=0` or `-tags netgo,osusergo -ldflags "-linkmode 'external' -extldflags '-static'"` but from what I have [read](https://mt165.co.uk/blog/static-link-go/) it seems that `CGO_ENABLED=0` is the preferred way of building a static binary.

The blog says:

> To use the libc functions for net and os/user, and still get a static binary (for containers): go build -ldflags "-linkmode 'external' -extldflags '-static'"

`net` and `os/user` are stdlib packages. As I understand, the above statement holds true for non-stdlib packages too that import C.

Is there a way we can verify that the C bindings are still linked when we compile with `CGO_ENABLED=0`? Maybe by trying to run checkpointctl in a container without glibc. If it works, then no issues at all. It's just that the way Go's documentation is worded implies to me that `CGO_ENABLED=0` skips compilation (and by extension, linking) of C code. There is a [GitHub issue](https://github.com/golang/go/issues/12860) about warning users when they compile Go code with files that `import "C"` if CGo is disabled, although it wasn't implemented.

rst0git:
> Is there a way we can verify that the C bindings are still linked when we compile with CGO_ENABLED=0? 

When the `CGO_ENABLED` is set to `0`, it tells the Go build system to completely disable the use of Cgo. As a result it builds a static binary without any C dependencies. The resulting executable is built from Go code that is entirely self-contained and doesn't rely on external C libraries.

In the case of `net`, `os/user`, and `containers/storage` they provide Go code that is used as an alternative when building without CGO (as Andrei pointed out above).

snprajwal:
> In the case of net, os/user, and containers/storage they provide Go code that is used as an alternative when building without CGO (as Andrei pointed out above).

Ah alright, in that case it should work. Thanks for the clarification!

snprajwal (approved):

PR #66, opened 2023-06-10
show: enable shorthand options

rst0git:
This patch enables the following shorthand options:

```
Usage:
  checkpointctl show [flags]

Flags:
  -A, --all          Display all additional information about the checkpoints
  -F, --full-paths   Display mounts with full paths
  -h, --help         help for show
  -m, --mounts       Print overview about mounts used in the checkpoints
  -s, --stats        Print checkpointing statistics if available
```

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -52,27 +52,31 @@ func setupShow() *cobra.Command {
 		false,
 		"Print checkpointing statistics if available",
 	)
-	flags.BoolVar(
+	flags.BoolVarP(
 		&stats,
 		"stats",
+		"s",
 		false,
 		"Print checkpointing statistics if available",
 	)
-	flags.BoolVar(
+	flags.BoolVarP(
 		&mounts,
 		"mounts",
+		"m",
 		false,
 		"Print overview about mounts used in the checkpoints",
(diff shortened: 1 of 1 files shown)


snprajwal:
Wouldn't some of these shorthands conflict with future additions? E.g. `-f` for `--full-paths` could be mistaken for `--fds`.

rst0git:
> Wouldn't some of these shorthands conflict with future additions? E.g. `-f` for `--full-paths` could be mistaken for `--fds`.

Good question. I expect there will be similar conflicts with future additions.
Should use `-A` and `-F` (capitals) for `--all` and `--full-paths`?

snprajwal:
That would work well

snprajwal (approved):

adrianreber:
I understand the reason for this PR, but I am still a bit undecided. Maybe we should wait a bit longer until we have more features before introducing shorthand options. Once we have more flags, maybe we can see better how to distribute the shorthand flags. Just an idea.

rst0git:
Closing this PR in favour of the changes introduced with https://github.com/checkpoint-restore/checkpointctl/pull/76

Issue #53, opened 2023-05-17
Add support for showing information for multiple checkpoints


rst0git:
The `checkpointctl show` command currently supports displaying the information for a single checkpoint file. However, it would be very useful for this command to be extended with support for multiple files (similar to other Unix commands).

Example:

```bash
checkpointctl show /var/lib/kubelet/checkpoints/*.tar
```

behouba:
Hi @rst0git, 
I would like to work on this.

rst0git:
It sounds like a good idea.

@adrianreber What do you think?

adrianreber:
> It sounds like a good idea.
> 
> @adrianreber What do you think?

Sounds good.

The more important thing from my point of view would be to focus on unpacking only the needed files. Multiple checkpoints can be something for later.

For multiple checkpoints I would at first say to not support the additional flags. Maybe just list all the checkpoints in a single table.

behouba:
Thank you, @adrianreber, for your suggestion. I will also start thinking about #51.

adrianreber:
> Thank you, @adrianreber, for your feedback. I will also start thinking about #51.

Ah, sorry. I got confused. I thought this was #51. Then my comment does not make so much sense.

I still think I would start with a simple show all checkpoints in one table. My use case would be to get a quick overview of all checkpoints.

Details are not so important in that case at first.

behouba:
Alright, I agree that it would be better to initially not support additional flags when handling multiple files. 
What do you think, @rst0git?

rst0git:
> I still think I would start with a simple show all checkpoints in one table. My use case would be to get a quick overview of all checkpoints.

I also agree, using one table with an overview of all checkpoints would be more appropriate.

Example:
```
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
|   CONTAINER    |             IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| gallant_austin | docker.io/library/nginx:latest | 5e61a1e0a308 | crun    | 2023-03-01T15:11:55+03:00 | Podman | 16.6 MiB   | 4.5 KiB           |
+----------------+--------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
[...]

PR #67, opened 2023-06-10
show: add support for listing multiple checkpoints

behouba:
This commit enhances the functionality of the "show" command by enabling it to list multiple checkpoint archives.

Examples:
```bash
$ls checkpoints/
nginx-checkpoint.tar.gz  ubuntu_multiuser.tar.gz

$checkpointctl nginx-checkpoint.tar.gz  ubuntu_multiuser.tar.gz
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+
|   CONTAINER    |              IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | IP | MAC | CHKPT SIZE | ROOT FS DIFF SIZE |
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+
| awesome_booth  | docker.io/library/ubuntu:latest | 695b77deb382 | crun    | 2023-03-08T08:45:33+03:00 | Podman |    |     | 2.8 MiB    | 309.0 KiB         |
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+
| gallant_austin | docker.io/library/nginx:latest  | 5e61a1e0a308 | crun    | 2023-03-01T15:11:55+03:00 | Podman |    |     | 16.6 MiB   | 4.5 KiB           |
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+

$checkpointctl show *.tar.gz
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+
|   CONTAINER    |              IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | IP | MAC | CHKPT SIZE | ROOT FS DIFF SIZE |
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+
| awesome_booth  | docker.io/library/ubuntu:latest | 695b77deb382 | crun    | 2023-03-08T08:45:33+03:00 | Podman |    |     | 2.8 MiB    | 309.0 KiB         |
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+
| gallant_austin | docker.io/library/nginx:latest  | 5e61a1e0a308 | crun    | 2023-03-01T15:11:55+03:00 | Podman |    |     | 16.6 MiB   | 4.5 KiB           |
+----------------+---------------------------------+--------------+---------+---------------------------+--------+----+-----+------------+-------------------+

```

Note that when listing multiple checkpoint files, additional flags are ignored.

Fixes: #53

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -102,55 +102,65 @@ func show(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("Cannot use --full-paths without --mounts/--all option")
 	}
 
-	input := args[0]
-	tar, err := os.Stat(input)
-	if err != nil {
-		return err
-	}
-	if !tar.Mode().IsRegular() {
-		return fmt.Errorf("input %s not a regular file", input)
-	}
+	tasks := make([]task, 0, len(args))
 
-	// Check if there is a checkpoint directory in the archive file
-	checkpointDirExists, err := isFileInArchive(input, metadata.CheckpointDirectory, true)
-	if err != nil {
-		return err
-	}
-
-	if !checkpointDirExists {
-		return fmt.Errorf("checkpoint directory is missing in the archive file: %s", input)
-	}
-
-	dir, err := os.MkdirTemp("", "checkpointctl")
-	if err != nil {
-		return err
-	}
-	defer func() {
-		if err := os.RemoveAll(dir); err != nil {
-			fmt.Fprintln(os.Stderr, err)
+	for _, input := range args {
+		tar, err := os.Stat(input)
+		if err != nil {
+			return err
 		}
-	}()
+		if !tar.Mode().IsRegular() {
+			return fmt.Errorf("input %s not a regular file", input)
+		}
 
-	// A list of files that need to be unarchived. The files need not be
-	// full paths. Even a substring of the file name is valid.
-	files := []string{metadata.SpecDumpFile, metadata.ConfigDumpFile}
+		// A list of files that need to be unarchived. The files need not be
+		// full paths. Even a substring of the file name is valid.
+		files := []string{metadata.SpecDumpFile, metadata.ConfigDumpFile}
 
-	if stats {
-		files = append(files, "stats-dump")
+		if stats {
+			files = append(files, "stats-dump")
+		}
+
+		if psTree {
+			files = append(
+				files,
+				filepath.Join(metadata.CheckpointDirectory, "pstree.img"),
+				// All core-*.img files
+				filepath.Join(metadata.CheckpointDirectory, "core-"),
+			)
+		}
+
+		// Check if there is a checkpoint directory in the archive file
+		checkpointDirExists, err := isFileInArchive(input, metadata.CheckpointDirectory, true)
+		if err != nil {
+			return err
+		}
+
+		if !checkpointDirExists {
+			return fmt.Errorf("checkpoint directory is missing in the archive file: %s", input)
+		}
+
+		dir, err := os.MkdirTemp("", "checkpointctl")
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err := os.RemoveAll(dir); err != nil {
+				fmt.Fprintln(os.Stderr, err)
+			}
+		}()
+
+		if err := untarFiles(input, dir, files); err != nil {
+			return err
+		}
+
+		tasks = append(tasks, task{checkpointFilePath: input, outputDir: dir})
 	}
 
-	if psTree {
-		files = append(
-			files,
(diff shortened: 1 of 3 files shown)


adrianreber:
> Note that when listing multiple checkpoint files, additional flags are ignored.

Not sure ignored is the correct behaviour. I think we should fail early.

@rst0git what do you think? Should we ignore flags or fail early?

snprajwal:
If we look at the descriptions for our flags, they all mention "checkpoints" in plural. E.g. https://github.com/checkpoint-restore/checkpointctl/blob/34bab9962e279cc38f1ad26aad814b64fe9f16b9/checkpointctl.go#L61-L65

I think it would make sense to neither ignore nor fail, but apply the flags to each checkpoint that we are displaying.

adrianreber:
> I think it would make sense to neither ignore nor fail, but apply the flags to each checkpoint that we are displaying.

We discussed this a bit in #53. If we want to output all the additional information for multiple checkpoints the output gets messy and not sure how useful it is. The table view is good if we want show the same data multiple times (like in this PR), but mixing different tables doesn't sound helpful. Maybe the tree view would also be better... Not sure.

One of the ideas in #53 was that if multiple checkpoint archives are specified then just give an overview of what is available and only in the single view support additional information.

As in other tickets discussed the best way to display the huge amount of data to the user is not yet clear. In the terminal the horizontal space is limited, but we can use as much vertical space as we want. So maybe we need to use the tree approach much more... Maybe the table just doesn't work.

rst0git:
I also think that it would be useful to apply flags (such as `--mounts`, `--stats`) to each checkpoint. I find this functionality useful when comparing the results of different checkpoints. However, we can implement this in another pull request.

> If we want to output all the additional information for multiple checkpoints the output gets messy and not sure how useful it is. The table view is good if we want show the same data multiple times (like in this PR), but mixing different tables doesn't sound helpful.

We can add an option `--output` to specify an output format (e.g., `json`, `yaml`) similar to `kubectl`.

checkpointctl.go
```diff
@@ -93,38 +93,48 @@ func show(cmd *cobra.Command, args []string) error {
...
 
-	input := args[0]
-	tar, err := os.Stat(input)
-	if err != nil {
-		return err
-	}
-	if !tar.Mode().IsRegular() {
-		return fmt.Errorf("input %s not a regular file", input)
-	}
+	tasks := make([]task, 0)
```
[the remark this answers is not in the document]
behouba: True

container.go
```diff
@@ -66,77 +66,117 @@ func getCRIOInfo(_ *metadata.ContainerConfig, specDump *spec.Spec) (*containerIn
...
 }
 
-func showContainerCheckpoint(checkpointDirectory, input string) error {
-	containerConfig, _, err := metadata.ReadContainerCheckpointConfigDump(checkpointDirectory)
-	if err != nil {
-		return err
+func showContainerCheckpoints(tasks []task) error {
+	// Return an error early when attempting to display multiple checkpoints with mount or stats flags
+	if (len(tasks) > 1) && (mounts || stats) {
+		return fmt.Errorf("displaying multiple checkpoints with mount or stats flags is not supported")
```
rst0git, line 72: There might be a few more flags added to checkpointctl, and it might be better to use a more generic error message instead of listing all flags (e.g., "displaying multiple checkpoints with _additional_ flags is not supported")

container.go
```diff
@@ -148,31 +188,13 @@ func renderCheckpoint(ci *containerInfo, containerConfig *metadata.ContainerConf
...
-		header = append(header, "IP")
-		row = append(row, ci.IP)
-	}
-	if ci.MAC != "" {
-		header = append(header, "MAC")
-		row = append(row, ci.MAC)
-	}
-
-	header = append(header, "CHKPT Size")
+	row = append(row, ci.IP)
```
adrianreber, line 192: Why are you display the MAC address and IP address unconditionally even if it is not set?

adrianreber: Right, that makes sense. Tables are just dumb :wink: or it least not helpful here.

I guess for listing multiple checkpoints your approach is correct, but for the single checkpoint view you should make it conditional.

behouba: Ok, I will try something :)

behouba: @adrianreber, I made changes to maintain the previous output format when displaying a single checkpoint. What do you think of this approach?

rst0git:
@behouba Would you be able to rebase this pull request?

PR #71, opened 2023-06-16
fix: remove redundant nodes in process tree

### LINKED ISSUE #70: Process tree output looks not totally correct.
Using `--ps-tree` on a container checkpoint I get following output:
```
counter
└── [1]  bash
    ├── [7]  bash
    ├── [7]  counter.py
    ├── [8]  bash
    └── [8]  tee
```
That doesn't look like what I expected. I would expect something like:
```
counter
└── [1]  bash
    ├── [7]  counter.py
    └── [8]  tee
```

@snprajwal PTAL

snprajwal:
The recursive logic used to process nodes in the process tree was accidentally appending child nodes twice. This has been fixed by removing the addition of the extra node at the same depth before processing the child.

Fixes #70

### DIFF
--- a/container.go
+++ b/container.go
@@ -241,7 +241,6 @@ func renderPsTree(psTree *crit.PsTree, containerName string) {
 	processNodes = func(tree treeprint.Tree, root *crit.PsTree) {
 		node := tree.AddMetaBranch(root.PId, root.Comm)
 		for _, child := range root.Children {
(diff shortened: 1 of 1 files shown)


snprajwal:
Nice, one more bug in the binary size checker. We should be comparing against the last non-merge commit, not the entire history of merge commits. I'll open a PR for this.

rst0git (approved):
LGTM

PR #72, opened 2023-06-16
fix(ci/check-size): use latest non-merge commit

snprajwal:
We were previously using the merge commit as the baseline, leading to all commits from the pull request being rebased, rather than just the latest commit.


adrianreber (approved):
Looks good.

@rst0git PTAL

rst0git (approved):

PR #73, opened 2023-06-18
Add container test with Podman

### LINKED ISSUE #68: Add integration test with podman
Currently, we only run tests with BATS. This does not let us test changes in real world scenarios with containers. An integration test where a container is checkpointed and then the archive used for testing is much more robust and bulletproof.

snprajwal:
This currently fails because I'm doing some container inception (trying to checkpoint a podman container running on a Fedora container in GH Actions). Not sure why CRIU is failing, and podman doesn't provide a verbose flag to log info about the checkpointing process.

Fixes #68

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -6,18 +6,19 @@ jobs:
   coverage:
     runs-on: ubuntu-latest
     steps:
-      - name: checkout
+      - name: Checkout
         uses: actions/checkout@v3
         with:
           # needed for codecov
           fetch-depth: 0
       - name: Install tools
         run: |
-          sudo apt-get install -qqy bats
           # Add PPA for CRIU
           sudo add-apt-repository ppa:criu/ppa
           sudo apt-get update
           sudo apt-get install -qqy criu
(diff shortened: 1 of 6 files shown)


snprajwal:
Looks like GitHub runners don't support checkpointing with podman. The workflows throw an error for `podman container checkpoint` saying `configured runtime does not support checkpoint/restore`. We might have to use an external bare metal runner or a VM for this (maybe the CirrusCI instances we use with CRIU?)

adrianreber:
> Looks like GitHub runners don't support checkpointing with podman. The workflows throw an error for `podman container checkpoint` saying `configured runtime does not support checkpoint/restore`. We might have to use an external bare metal runner or a VM for this (maybe the CirrusCI instances we use with CRIU?)

The version of crun probably has not been compiled with CRIU support. Just switching to runc should already fix this.

snprajwal:
Using runc lets us checkpoint, but the `stats-dump` file is missing for some reason. Not sure why. The logs are available here: https://github.com/snprajwal/checkpointctl/actions/runs/5477217792/jobs/9975920598#step:6:50

PR #74, opened 2023-06-18
feat: display file descriptors

snprajwal:
The file descriptors opened by the processes in the checkpoint can now be viewed with the `--fds` flag. This currently displays the file type, FD, and the path.

An example output for a checkpoint of httpd container (provided by @rst0git):

```
test
├── Image: docker.io/library/httpd:2.4
├── ID: 41ddbcab288ab3d2d119896aa96187b474bbe3084640b6559b0edc0a41f12277
├── Runtime: crun
├── Created: 2023-07-06T09:08:34+01:00
├── Engine: Podman
├── Checkpoint Size: 5.2 MiB
├── Root Fs Diff Size: 2.0 KiB
└── Process tree
    └── [1]  httpd
        ├── [3]  httpd
        │   ├── [REG 0]  /dev/null
        │   ├── [PIPE 1]  pipe[793149]
        │   ├── [PIPE 2]  pipe[793150]
        │   ├── [INETSK 3]  INETSK.36
        │   ├── [INETSK 4]  INETSK.37
        │   ├── [PIPE 5]  pipe[794937]
        │   ├── [PIPE 6]  pipe[794937]
        │   ├── [PIPE 7]  pipe[793149]
        │   ├── [EVENTPOLL 8]  EVENTPOLL.44
        │   ├── [PIPE 9]  pipe[796833]
        │   ├── [PIPE 10]  pipe[796833]
        │   ├── [cwd]  /usr/local/apache2
        │   └── [root]  /
        ├── [4]  httpd
        │   ├── [REG 0]  /dev/null
        │   ├── [PIPE 1]  pipe[793149]
        │   ├── [PIPE 2]  pipe[793150]
        │   ├── [INETSK 3]  INETSK.36
        │   ├── [INETSK 4]  INETSK.37
        │   ├── [PIPE 5]  pipe[794937]
        │   ├── [PIPE 6]  pipe[794937]
        │   ├── [PIPE 7]  pipe[793149]
        │   ├── [EVENTPOLL 8]  EVENTPOLL.49
        │   ├── [PIPE 9]  pipe[789308]
        │   ├── [PIPE 10]  pipe[789308]
        │   ├── [cwd]  /usr/local/apache2
        │   └── [root]  /
        ├── [5]  httpd
        │   ├── [REG 0]  /dev/null
        │   ├── [PIPE 1]  pipe[793149]
        │   ├── [PIPE 2]  pipe[793150]
        │   ├── [INETSK 3]  INETSK.36
        │   ├── [INETSK 4]  INETSK.37
        │   ├── [PIPE 5]  pipe[794937]
        │   ├── [PIPE 6]  pipe[794937]
        │   ├── [PIPE 7]  pipe[793149]
        │   ├── [EVENTPOLL 8]  EVENTPOLL.54
        │   ├── [PIPE 9]  pipe[792449]
        │   ├── [PIPE 10]  pipe[792449]
        │   ├── [cwd]  /usr/local/apache2
        │   └── [root]  /
        ├── [REG 0]  /dev/null
        ├── [PIPE 1]  pipe[793149]
        ├── [PIPE 2]  pipe[793150]
        ├── [INETSK 3]  INETSK.36
        ├── [INETSK 4]  INETSK.37
        ├── [PIPE 5]  pipe[794937]
        ├── [PIPE 6]  pipe[794937]
        ├── [PIPE 7]  pipe[793149]
        ├── [cwd]  /usr/local/apache2
        └── [root]  /
```


checkpointctl.go
```diff
@@ -19,6 +19,7 @@ var (
 	mounts    bool
 	fullPaths bool
 	psTree    bool
+	fds       bool
```
rst0git, line 22: What about using `--files` (instead of `--fds`) for better readability?

checkpointctl.go
```diff
@@ -93,19 +94,25 @@ func setupInspect() *cobra.Command {
...
+		"files",
+		false,
+		"Display an overview of files used in the container checkpoint",
+	)
 	flags.BoolVar(
 		&showAll,
 		"all",
 		false,
-		"Show all information about container checkpoint",
+		"Show all information about the container checkpoint",
```
rst0git, line 115: We intentionally don't use "the" here because it would suggest that this option is designed to show information about a particular checkpoint. By omitting "the", we convey that `--all` is capable of providing information about any container checkpoint, which is aligned with the functionality supporting multiple checkpoints, e.g., `checkpointctl inspect --all *.tar.gz`.

snprajwal: Got it. In that case, "checkpoints" would make more sense.

checkpointctl.go
```diff
@@ -77,12 +76,6 @@ func setupInspect() *cobra.Command {
 	}
 	flags := cmd.Flags()
 
-	flags.BoolVar(
-		&stats,
-		"print-stats",
-		false,
-		"Display checkpoint statistics",
-	)
```
rst0git, line 85: This change is not directly related to the feature displaying file descriptors. Would it be possible to create a new pull request?

snprajwal: Righto, I'll move both the fix commits to a separate PR

test/checkpointctl.bats
```diff
@@ -246,7 +246,7 @@ function teardown() {
 	checkpointctl inspect "$TEST_TMP_DIR2"/test.tar --stats
 	[ "$status" -eq 0 ]
 	[[ ${lines[8]} == *"CRIU dump statistics"* ]]
-	[[ ${lines[12]} == *"Memwrite Time"* ]]
+	[[ ${lines[12]} == *"Memwrite time"* ]]
```
rst0git, line 249: Could you create a new pull request for this patch (fix: capitalisation in tree output)?

checkpointctl.go
```diff
@@ -93,19 +94,25 @@ func setupInspect() *cobra.Command {
...
 		&psTree,
 		"ps-tree",
 		false,
 		"Display an overview of processes in the container checkpoint",
 	)
+	flags.BoolVar(
+		&files,
+		"files",
+		false,
+		"Display an overview of files used in the container checkpoint",
```
rst0git, line 109: Considering the output of this option, the following description might be more appropriate: "Display a list of open file descriptors for each process in the container checkpoint"


For example, here is the output for a checkpoint of httpd container:
```
test
├── Image: docker.io/library/httpd:2.4
├── ID: 41ddbcab288ab3d2d119896aa96187b474bbe3084640b6559b0edc0a41f12277
├── Runtime: crun
├── Created: 2023-07-06T09:08:34+01:00
├── Engine: Podman
├── Checkpoint Size: 5.2 MiB
├── Root Fs Diff Size: 2.0 KiB
└── Process tree
    └── [1]  httpd
        ├── [3]  httpd
        │   ├── [REG 0]  /dev/null
        │   ├── [PIPE 1]  pipe[793149]
        │   ├── [PIPE 2]  pipe[793150]
        │   ├── [INETSK 3]  INETSK.36
        │   ├── [INETSK 4]  INETSK.37
        │   ├── [PIPE 5]  pipe[794937]
        │   ├── [PIPE 6]  pipe[794937]
        │   ├── [PIPE 7]  pipe[793149]
        │   ├── [EVENTPOLL 8]  EVENTPOLL.44
        │   ├── [PIPE 9]  pipe[796833]
        │   ├── [PIPE 10]  pipe[796833]
        │   ├── [cwd]  /usr/local/apache2
        │   └── [root]  /
        ├── [4]  httpd
        │   ├── [REG 0]  /dev/null
        │   ├── [PIPE 1]  pipe[793149]
        │   ├── [PIPE 2]  pipe[793150]
        │   ├── [INETSK 3]  INETSK.36
        │   ├── [INETSK 4]  INETSK.37
        │   ├── [PIPE 5]  pipe[794937]
        │   ├── [PIPE 6]  pipe[794937]
        │   ├── [PIPE 7]  pipe[793149]
        │   ├── [EVENTPOLL 8]  EVENTPOLL.49
        │   ├── [PIPE 9]  pipe[789308]
        │   ├── [PIPE 10]  pipe[789308]
        │   ├── [cwd]  /usr/local/apache2
        │   └── [root]  /
        ├── [5]  httpd
        │   ├── [REG 0]  /dev/null
        │   ├── [PIPE 1]  pipe[793149]
        │   ├── [PIPE 2]  pipe[793150]
        │   ├── [INETSK 3]  INETSK.36
        │   ├── [INETSK 4]  INETSK.37
        │   ├── [PIPE 5]  pipe[794937]
        │   ├── [PIPE 6]  pipe[794937]
        │   ├── [PIPE 7]  pipe[793149]
        │   ├── [EVENTPOLL 8]  EVENTPOLL.54
        │   ├── [PIPE 9]  pipe[792449]
        │   ├── [PIPE 10]  pipe[792449]
        │   ├── [cwd]  /usr/local/apache2
        │   └── [root]  /
        ├── [REG 0]  /dev/null
        ├── [PIPE 1]  pipe[793149]
        ├── [PIPE 2]  pipe[793150]
        ├── [INETSK 3]  INETSK.36
        ├── [INETSK 4]  INETSK.37
        ├── [PIPE 5]  pipe[794937]
        ├── [PIPE 6]  pipe[794937]
        ├── [PIPE 7]  pipe[793149]
        ├── [cwd]  /usr/local/apache2
        └── [root]  /

```

rst0git (approved):
LGTM

adrianreber:
The binary size check does not seem to work any more. I just see one run of it. No comparison. 

Can you add an example of what this looks like in the PR description?

rst0git:
> The binary size check does not seem to work any more. I just see one run of it. No comparison.

The output looks correct. We see only one run because there is only 1 commit in the pull request.

```
Changes from 027219c102cc9236b72ca5ba1a401d81e2dc7694 to 027219c102cc9236b72ca5ba1a401d81e2dc7694:
Rebasing (2/2)
Executing: test/check-size.sh
go build -o checkpointctl -ldflags "-X main.name=checkpointctl -X main.version=0.1.0"
BINARY SIZE (ec8add6): 9998896
No previous size present, storin
```

snprajwal:
The binary size workflow should be working now, I rebased this branch against main.

snprajwal:
I forgot to do a nil pointer check while searching for the process node in the tree. Updated the commit with those changes now.

tree.go
```diff
@@ -100,16 +111,35 @@ func addDumpStatsToTree(tree treeprint.Tree, dumpStats *images.DumpStatsEntry) {
 }
 
 func addPsTreeToTree(tree treeprint.Tree, psTree *crit.PsTree) {
-	// processNodes is a recursive function to create
-	// a new branch for each process and add its child
-	// processes as child nodes of the branch.
+	/* processNodes is a recursive function to create
+	a new branch for each process and add its child
+	processes as child nodes of the branch. */
```
rst0git, line 116: The convention for comments in Go is to use C++-style `//` line comments.
https://go.dev/doc/effective_go#commentary

tree.go
```diff
@@ -100,16 +111,35 @@ func addDumpStatsToTree(tree treeprint.Tree, dumpStats *images.DumpStatsEntry) {
...
+		node = tree.FindByMeta(fd.PId)
+		if node == nil {
+			/* If the PID is not present in the process tree,
+			it is either invalid, or has been pruned while
+			building the process tree. In the first case,
+			the error will be handled by the process tree
+			builder. In the second case, we simply omit adding
+			those FDs. */
+			continue
+		}
```
rst0git, line 140: Hmm, in ExploreFds() we have the following code:

```Go
pID := process.GetPid()
// Get file with object IDs
idsImg, err := getImg(filepath.Join(c.inputDirPath, fmt.Sprintf("ids-%d.img", pID)), &criu_core.TaskKobjIdsEntry{})
if err != nil {
	return nil, err
}
filesID := idsImg.Entries[0].Message.(*criu_core.TaskKobjIdsEntry).GetFilesId()
// Get open file descriptors
fdInfoImg, err := getImg(filepath.Join(c.inputDirPath, fmt.Sprintf("fdinfo-%d.img", filesID)), &fdinfo.FdinfoEntry{})
if err != nil {
	return nil, err
}

fdEntry := Fd{PId: pID}
```

Why `tree.FindByMeta(fd.PId)` would return `nil`?

tree.go
snprajwal: Since it was a comment explaining an internal function, I thought it would make more sense to use a multi-line comment to differentiate it from doc comment. Will change this.

tree.go
snprajwal: This is twofold. Firstly, in case we have an invalid checkpoint, and still successfully make it to this stage, if `tree.FindByMeta()` does not find the node and returns nil, it will cause a panic due to nil pointer dereference. This can be handled more gracefully by checking for the nil value.
Secondly, this is related to the change in #87 - if we prune the process tree for a particular PID, the other PID nodes will not be present. Hence, for those PID's file descriptors, `tree.FindByMeta()` will return nil.

rst0git: >   in case we have an invalid checkpoint, and still successfully make it to this stage, if tree.FindByMeta() does not find the node and returns nil, it will cause a panic due to nil pointer dereference. This can be handled more gracefully by checking for the nil value.

If CRIU fails, the container engine (CRI-O/Podman) is not going to create a checkpoint archive. It might be better to show an appropriate error message instead of silently ignoring this.

> Secondly, this is related to the change in https://github.com/checkpoint-restore/checkpointctl/pull/87 - if we prune the process tree for a particular PID, the other PID nodes will not be present. Hence, for those PID's file descriptors, tree.FindByMeta() will return nil.

To enable process tree output for specified PID we can introduce a similar change in #87. For example, it might be more efficient to pass the value specified with `--pid` as a parameter to `addFdsToTree()` and use something like the following:
```Go
if pid != fd.PId {
	continue
}
```

snprajwal: If you prefer, I can revert this change, and wait until this PR is merged. Since the change is directly related to #87, I can introduce this there after rebasing. What do you think?

rst0git: It would make sense to introduce this change in #87 after rebase.

PR #76, opened 2023-06-20
show: extend the output format with a tree view

behouba:
This PR aims to add the tree view output format to the `show` command. A new format flag is introduced to allow the selection of the desired output format. The tree view and table view have been moved to separate files for improved code organization.

I am considering designing a struct that will encompasses all the necessary data required by the `show` command. This will simplify future extensions, particularly for a JSON output format later.


table.go
```diff
@@ -0,0 +1,208 @@
...
+	processNodes = func(tree treeprint.Tree, root *crit.PsTree) {
+		node := tree.AddMetaBranch(root.PId, root.Comm)
+		for _, child := range root.Children {
+			processNodes(node, child)
+		}
+	}
+
+	processNodes(tree, psTree)
+
+	return tree
```
snprajwal, line 207: I think it would make sense to move the printing logic here to keep it consistent with the other rendering functions

tree.go
```diff
@@ -0,0 +1,125 @@
...
+		if psTree {
+			// The image files reside in a subdirectory called "checkpoint"
+			c := crit.New("", "", filepath.Join(task.outputDir, "checkpoint"), false, false)
+			// Get process tree with CRIT
+			psTree, err := c.ExplorePs()
+			if err != nil {
+				return fmt.Errorf("failed to get process tree: %w", err)
+			}
+
+			pTree := tree.AddBranch("Processes")
```
snprajwal, line 105: It would be better to name this `psTreeNode` instead. `pTree` and `psTree` would be too easy to mix up.

tree.go
```diff
@@ -0,0 +1,129 @@
...
+		}()))
+	}
+}
+
+func addDumpStatsToTree(tree treeprint.Tree, dumpStats *images.DumpStatsEntry) {
+	statsTree := tree.AddBranch("CRIU dump statistics")
+	statsTree.AddBranch(fmt.Sprintf("Freezing Time: %d us", dumpStats.GetFreezingTime()))
+	statsTree.AddBranch(fmt.Sprintf("Frozen Time: %d us", dumpStats.GetFrozenTime()))
+	statsTree.AddBranch(fmt.Sprintf("Memdump Time: %d us", dumpStats.GetMemdumpTime()))
+	statsTree.AddBranch(fmt.Sprintf("FMemwrite Time: %d us", dumpStats.GetMemwriteTime()))
```
rst0git, line 111: Should this be "Memwrite Time" (without `F`)?

behouba: Yes, I is a typos 😅. I will fix that. Thank you you pointing this out.

rst0git (commented):
Would it make sense to always use `--all` and `--full-paths` with `--format tree` / `--format json`?

container.go
```diff
@@ -68,126 +64,10 @@ func getCRIOInfo(_ *metadata.ContainerConfig, specDump *spec.Spec) (*containerIn
 }
 
 func showContainerCheckpoints(tasks []task) error {
-	// Return an error early when attempting to display multiple checkpoints with additional flags
-	if (len(tasks) > 1) && (mounts || stats || psTree) {
-		return fmt.Errorf("displaying multiple checkpoints with additional flags is not supported")
+	if format == "table" {
+		return renderTableView(tasks)
```
rst0git, line 68: It might be good to use a `switch` statement here and show an error when the value of `format` is invalid.

checkpointctl.go
```diff
@@ -84,6 +85,12 @@ func setupShow() *cobra.Command {
 		false,
 		"Display all additional information about the checkpoints",
 	)
+	flags.StringVar(
+		&format,
+		"format",
+		"tree",
+		"Format the output using one of the following views: table, tree, or JSON (default: \"table\").",
```
snprajwal, line 92:
```suggestion
		"Format the output using one of the following views: table, tree, or JSON (default: \"tree\").",
```

behouba:
> Would it make sense to always use `--all` and `--full-paths` with `--format tree` / `--format json`?

It makes sense, given that the tree format with "--all" and "--full-path" won't cause the terminal screen to be overcrowded in general.

snprajwal:
Given that tree view is the default, and JSON is better than the table view, do we even want to retain it? It might make sense to just get rid of it if we don't plan on using it going forward

behouba:
> Given that tree view is the default, and JSON is better than the table view, do we even want to retain it? It might make sense to just get rid of it if we don't plan on using it going forward.

While I don't have a strong opinion on this, it's an interesting question.

rst0git:
> Given that tree view is the default, and JSON is better than the table view, do we even want to retain it? It might make sense to just get rid of it if we don't plan on using it going forward.

It would make sense to keep the *table* as default because it shows a brief summary of the content of a checkpoint
and allows to display an overview of multiple checkpoints.

Example:
```
Displaying container checkpoint data from /tmp/checkpoint.tar.gz

+-----------+----------------------------------+--------------+---------+---------------------------+--------+------------+
| CONTAINER |              IMAGE               |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE |
+-----------+----------------------------------+--------------+---------+---------------------------+--------+------------+
| looper    | docker.io/library/busybox:latest | a69f3e73ec89 | crun    | 2023-06-25T11:00:18+01:00 | Podman | 154.5 KiB  |
+-----------+----------------------------------+--------------+---------+---------------------------+--------+------------+
```

In comparison, when displaying *tree* and *json* format with `--all` and `--full-path` (as discussed above), checkpointctl would provide a complete information about a checkpoint archive.

Example:
```
Displaying container checkpoint tree view from /tmp/checkpoint.tar.gz

looper
├── Image: docker.io/library/busybox:latest
├── ID: a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319
├── Runtime: crun
├── Created: 2023-06-25T11:00:18+01:00
├── Engine: Podman
├── CHKPT Size: 154.5 KiB
├── Overview of Mounts
│   ├── Destination: /proc
│   │   ├── Type: proc
│   │   └── Source: proc
│   ├── Destination: /dev
│   │   ├── Type: tmpfs
│   │   └── Source: tmpfs
│   ├── Destination: /sys
│   │   ├── Type: sysfs
│   │   └── Source: sysfs
│   ├── Destination: /dev/pts
│   │   ├── Type: devpts
│   │   └── Source: devpts
│   ├── Destination: /dev/mqueue
│   │   ├── Type: mqueue
│   │   └── Source: mqueue
│   ├── Destination: /etc/hosts
│   │   ├── Type: bind
│   │   └── Source: /run/containers/storage/overlay-containers/a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319/userdata/hosts
│   ├── Destination: /dev/shm
│   │   ├── Type: bind
│   │   └── Source: /var/lib/containers/storage/overlay-containers/a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319/userdata/shm
│   ├── Destination: /run/.containerenv
│   │   ├── Type: bind
│   │   └── Source: /run/containers/storage/overlay-containers/a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319/userdata/.containerenv
│   ├── Destination: /run/secrets
│   │   ├── Type: bind
│   │   └── Source: /run/containers/storage/overlay-containers/a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319/userdata/run/secrets
│   ├── Destination: /etc/hostname
│   │   ├── Type: bind
│   │   └── Source: /run/containers/storage/overlay-containers/a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319/userdata/hostname
│   ├── Destination: /etc/resolv.conf
│   │   ├── Type: bind
│   │   └── Source: /run/containers/storage/overlay-containers/a69f3e73ec89d6d3c4999d00407a2cca96df212163da4e088107b75225821319/userdata/resolv.conf
│   └── Destination: /sys/fs/cgroup
│       ├── Type: cgroup
│       └── Source: cgroup
├── CRIU dump statistics
│   ├── Freezing Time: 102175 us
│   ├── Frozen Time: 143805 us
│   ├── Memdump Time: 1808 us
│   ├── Memwrite Time: 184 us
│   ├── Pages Scanned: 1195 us
│   └── Pages Written: 34 us
└── Process tree
    └── [1]  sh
```

@adrianreber What do you think?

checkpointctl.go
```diff
@@ -84,6 +85,12 @@ func setupShow() *cobra.Command {
 		false,
 		"Display all additional information about the checkpoints",
 	)
+	flags.StringVar(
+		&format,
+		"format",
+		"table",
+		"Format the output using one of the following views: table, tree, or json (default: \"tree\").",
```
snprajwal, line 92: I had suggested changing the default to "tree" since we were previously using that. If we're going to use "table" as the default, we should change it here too.

behouba: Thanks @snprajwal, I forgot to update the description.

adrianreber:
@rst0git and I talked about this a bit.

Our idea was to keep the current table as is if using the the `show` sub-command. The `show` sub-command shouldn't have any parameters and is only used for a quick overview of the checkpoints in the table. Should work with multiple checkpoints.

In addition to `show` there should be an `inspect` sub-command which defaults to the tree view and has the all the additionally introduced parameter to allow the user to select which details to include in the tree view.

Maybe we should include a hint in the description of `show` that `inspect` has many more options.

The introduction of the new sub-command is motivated that the output of `show` should not change based on the used parameters. With a new command like `inspect` the format would always be a tree as default.

behouba:
> With a new command like `inspect` the format would always be a tree as default.

@adrianreber, @rst0git should the `inspect` sub command support displaying table formats in addition to the default tree format upcoming JSON format.

adrianreber:
> > With a new command like `inspect` the format would always be a tree as default.
> 
> @adrianreber, @rst0git should the `inspect` sub command support displaying table formats in addition to the default tree format upcoming JSON format.

From my point of view it should not.

checkpointctl.go
```diff
@@ -84,83 +114,110 @@ func setupShow() *cobra.Command {
 		false,
 		"Display all additional information about the checkpoints",
 	)
+	flags.StringVar(
+		&format,
+		"format",
+		"tree",
+		"Format the output using a tree or json format",
```
snprajwal, line 121: I think something like this would be better in case we decide to support some other serialisable format later on.
```suggestion
		"Specify the output format [tree (default), json]",
```

snprajwal:
Regarding serialising to JSON, we have two ways to do it:

- Create a custom struct that utilises CRIT's structs for the fields, and marshal that through Go's `encoding/json`
- We can use a library to serialise the tree into a JSON and vice versa. The benefit of this is that we can dynamically generate the JSON, independent of the structure of the tree. There would be no struct definitions needed for us to use the representation.
 
I can hack the library together in a couple of days, but I would like to know if you guys think this approach is any more useful that the first one.

behouba:
> * We can use a library to serialise the tree into a JSON and vice versa. The benefit of this is that we can dynamically generate the JSON, independent of the structure of the tree. There would be no struct definitions needed for us to use the representation.

I like this option as well. Because it will provide some consistency with the table and tree format.

checkpointctl.go
```diff
@@ -4,15 +4,23 @@ package main
...
 
 var (
-	name    string
-	version string
+	name      string
+	version   string
+	format    string
+	stats     bool
+	mounts    bool
+	fullPaths bool
```
rst0git, line 21: Since `checkpointctl inspect` is using a more compact format to display checkpoint information (i.e., `tree` or `json`), I think we could remove the `--full-paths` option.

checkpointctl.go
```diff
@@ -46,51 +58,166 @@ func setupShow() *cobra.Command {
...
+		&stats,
+		"print-stats",
+		false,
+		"Print checkpointing statistics if available",
+	)
+	flags.BoolVar(
+		&stats,
+		"stats",
+		false,
+		"Print checkpointing statistics if available",
```
rst0git: "Display checkpoint statistics"

checkpointctl.go
```diff
@@ -46,51 +58,166 @@ func setupShow() *cobra.Command {
...
+	)
+	flags.BoolVar(
+		&mounts,
+		"mounts",
+		false,
+		"Print overview about mounts used in the checkpoints",
```
rst0git: "Display an overview of mounts used in the checkpoint"

checkpointctl.go
```diff
@@ -46,51 +58,166 @@ func setupShow() *cobra.Command {
...
+	)
+	flags.BoolVar(
+		&psTree,
+		"ps-tree",
+		false,
+		"Print overview about the process tree in the checkpoints",
```
rst0git: "Display an overview of processes in the container checkpoint"

checkpointctl.go
```diff
@@ -46,51 +58,166 @@ func setupShow() *cobra.Command {
...
+		&fullPaths,
+		"full-paths",
+		false,
+		"Display mounts with full paths",
+	)
+	flags.BoolVar(
+		&showAll,
+		"all",
+		false,
+		"Display all additional information about the checkpoints",
```
rst0git, line 115: "Show all information about container checkpoint"

checkpointctl.go
```diff
@@ -111,7 +111,7 @@ func setupInspect() *cobra.Command {
 		&format,
 		"format",
 		"tree",
-		"Format the output using a tree or json format",
+		"Specify the output format [tree (default), json]",
```
rst0git, line 114: Specifying "(default)" here is unnecessary.

```
./checkpointctl inspect --help
Display low-level information about a container checkpoint

Usage:
  checkpointctl inspect [flags]

Flags:
      --all             Show all information about container checkpoint
      --format string   Specify the output format [tree (default), json] (default "tree")
...
```

test/checkpointctl.bats
```diff
@@ -263,16 +262,14 @@ function teardown() {
 	[[ ${lines[10]} == *"proc"* ]]
 }
 
-@test "Run checkpointctl inspect with tar file and --mounts and --full-paths and valid spec.dump" {
+@test "Run checkpointctl inspect with tar file and --ps-tree and missing pstree.img" {
 	cp data/config.dump "$TEST_TMP_DIR1"
 	cp data/spec.dump "$TEST_TMP_DIR1"
 	mkdir "$TEST_TMP_DIR1"/checkpoint
 	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/test.tar . )
-	checkpointctl inspect "$TEST_TMP_DIR2"/test.tar --mounts --full-paths
```
rst0git: Would it be possible to apply these changes in the previous commit (test: add tests for inspect sub-command)? This would ensure that `make test` will pass after each commit and make it easier to review the pull request.

It would be good to mention why `--full-paths` has been removed in the commit message introducing the `inspect` command (inspect: add new sub-command inspect).

tree.go
```diff
@@ -0,0 +1,130 @@
...
+	tree.AddBranch(fmt.Sprintf("Engine: %s", ci.Engine))
+
+	if ci.IP != "" {
+		tree.AddBranch(fmt.Sprintf("IP: %s", ci.IP))
+	}
+	if ci.MAC != "" {
+		tree.AddBranch(fmt.Sprintf("MAC: %s", ci.MAC))
+	}
+
+	tree.AddBranch(fmt.Sprintf("CHKPT Size: %s", metadata.ByteToString(archiveSizes.checkpointSize)))
```
rst0git, line 72: I believe the abbreviation "CHKPT Size" was initially used to fit within the table view. However, with tree / json view perhaps with can use "Checkpoint Size" instead?

tree.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if archiveSizes.rootFsDiffTarSize != 0 {
+		tree.AddBranch(fmt.Sprintf("Root Fs Diff Size: %s", metadata.ByteToString(archiveSizes.rootFsDiffTarSize)))
```
rst0git, line 75: "Rootfs diff size:"

rst0git (approved):

PR #82, opened 2023-07-06
chore: remove extra deprecated flag

snprajwal:
The `print-stats` flag was deprecated in `checkpointctl show`, and need not be introduced as a legacy flag in `checkpointctl inspect`.


rst0git (approved):
LGTM

PR #84, opened 2023-07-06
ci: use separate job to set baseline binary size

snprajwal:
Sometimes our own intelligence works against us +_+ better to keep it simple


snprajwal:
Nvm, the CI was not working on the other PR due to the branch being outdated against main

PR #87, opened 2023-07-10
feat: filter by PID in process tree

### LINKED ISSUE #86: Prune process tree by PID if specified
Currently, the `--ps-tree` and `--files` flags print the entire process tree, along with the associated information. These flags can optionally use a PID, and prune the tree to show just that process and its children. This allows for reduced noise when the process that needs to be inspected is already known. E.g. the PID of a known process can be retrieved with `--ps-tree`, and the finer details can be printed specifically for that process with `checkpointctl inspect --files --sk --pid <PID>`. A short form `-p` can also be provided.

snprajwal:
Currently, the entire process tree is rendered in the tree view. This allows for the tree to be pruned for a specific process using the `--pid` or `-p` flag.

Fixes #86

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -17,6 +17,7 @@ var (
 	format  string
 	stats   bool
 	mounts  bool
+	pID     uint32
 	psTree  bool
 	files   bool
 	showAll bool
@@ -89,6 +90,13 @@ func setupInspect() *cobra.Command {
 		false,
 		"Display an overview of mounts used in the container checkpoint",
 	)
+	flags.Uint32VarP(
+		&pID,
+		"pid",
+		"p",
+		0,
+		"Display the process tree of a specific PID",
+	)
 	flags.BoolVar(
 		&psTree,
 		"ps-tree",
@@ -129,6 +137,11 @@ func inspect(cmd *cobra.Command, args []string) error {
 
 	if stats {
 		requiredFiles = append(requiredFiles, "stats-dump")
+	}
+
+	if pID != 0 {
+		// Enable displaying process tree if the PID filter is passed.
+		psTree = true
 	}
 
 	if files {
--- a/tree.go
+++ b/tree.go
@@ -37,7 +37,9 @@ func renderTreeView(tasks []task) error {
 				return fmt.Errorf("failed to get process tree: %w", err)
 			}
 
-			addPsTreeToTree(tree, psTree)
+			if err = addPsTreeToTree(tree, psTree); err != nil {
+				return fmt.Errorf("failed to get process tree: %w", err)
+			}
 		}
 
 		if files {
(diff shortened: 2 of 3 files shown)


tree.go
```diff
@@ -110,10 +112,33 @@ func addDumpStatsToTree(tree treeprint.Tree, dumpStats *stats_pb.DumpStatsEntry)
...
+		ps := dfs(psTree)
+		if ps == nil {
+			return fmt.Errorf("no process with PID %d", pID)
+		}
+		psRoot = ps
+	}
+
+	/* processNodes is a recursive function to create
+	a new branch for each process and add its child
+	processes as child nodes of the branch. */
```
rst0git, line 141: For consistency, it might be better to use C++-style comment here (`//`).

snprajwal: My bad, I carried the change over from the previous PR. This wasn't supposed to be there.

tree.go
```diff
@@ -110,10 +112,33 @@ func addDumpStatsToTree(tree treeprint.Tree, dumpStats *stats_pb.DumpStatsEntry)
...
+			}
+			for _, child := range root.Children {
+				if ps := dfs(child); ps != nil {
+					return ps
+				}
+			}
+			return nil
+		}
+
+		ps := dfs(psTree)
```
rst0git, line 132: Do we need to check if `dfs == nil`?

snprajwal: If you're asking about the nil check for line 132, we're doing it in the next line and reporting an error if we don't find any process with the specified PID.

tree.go
```diff
@@ -37,7 +37,9 @@ func renderTreeView(tasks []task) error {
 				return fmt.Errorf("failed to get process tree: %w", err)
 			}
 
-			addPsTreeToTree(tree, psTree)
+			if err = addPsTreeToTree(tree, psTree); err != nil {
+				return fmt.Errorf("failed to get process tree: %w", err)
+			}
 		}
 
 		if files {
```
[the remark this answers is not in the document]
snprajwal: Should be fixed now :eyes:

checkpointctl.go
```diff
@@ -89,6 +90,13 @@ func setupInspect() *cobra.Command {
 		false,
 		"Display an overview of mounts used in the container checkpoint",
 	)
+	flags.Uint32VarP(
+		&pID,
+		"pid",
+		"p",
+		0,
+		"Specify a particular process in the container checkpoint",
+	)
```
rst0git, line 99: Should we automatically set `psTree = true` when the `--pid` options is used?

snprajwal: :thinking: I don't think we should, because the `--pid` option is not exclusive to `--ps-tree`. It works with the child flags too, like `--files`. If `--pid` is passed without any flag that generates a process tree, we should do nothing, IMO.

checkpointctl.go
```diff
@@ -89,6 +90,13 @@ func setupInspect() *cobra.Command {
 		false,
 		"Display an overview of mounts used in the container checkpoint",
 	)
+	flags.Uint32VarP(
+		&pID,
+		"pid",
+		"p",
+		0,
+		"Specify a particular process in the container checkpoint",
```
rst0git, line 98: What about something like "Display the process tree of a specific PID"?

tree.go
```diff
@@ -110,7 +112,30 @@ func addDumpStatsToTree(tree treeprint.Tree, dumpStats *stats_pb.DumpStatsEntry)
...
+				if ps := dfs(child); ps != nil {
+					return ps
+				}
+			}
+			return nil
+		}
+
+		ps := dfs(psTree)
+		if ps == nil {
+			return fmt.Errorf("no process with PID %d", pID)
```
rst0git, line 134: Would it make sense to extend this message to also show the available PIDs?

snprajwal: It would work well for smaller checkpoints, but I'm unsure about how we can manage the output for bigger process trees. Either way, wouldn't it be trivial for the user to run `checkpointctl inspect --ps-tree` to see the available PIDs before using the PID flag?

checkpointctl.go
rst0git: Sure with `--ps-tree` and `--files` we generate a process tree, while `--pid <PID>` can be used as a filter. It would be easier to use when the default behaviour is to show a process tree without explicitly specifying `--ps-tree`.
```
checkpointctl inspect --pid 5 /tmp/checkpoint.tar.gz
checkpointctl inspect --files --pid 5 /tmp/checkpoint.tar.gz
```

tree.go
rst0git: It makes sense. Should we add something like the following to the error message?
> Use `checkpointctl inspect --ps-tree` to list available PIDs

rst0git (approved):

snprajwal:
cc @adrianreber PTAL?

PR #69, opened 2023-06-15
Proof of concept on the usage of `crit` features for reading memory pages

behouba:
This proof of concept introduces the following new flags:
- pid: to specify the PID of a process to display
- ps-args: to display command line arguments of a process
- ps-env-vars: to display environment variables of a process
- ps-memory-pages: to display the content of memory pages in hexadecimal and ascii format (similar to hexdump)

## Example output

Information about the new flags:

```bash
$ ./checkpointctl show --help
Show information about available checkpoints

Usage:
  checkpointctl show [flags]

Flags:
      --all               Display all additional information about the checkpoints
      --full-paths        Display mounts with full paths
  -h, --help              help for show
      --mounts            Print overview about mounts used in the checkpoints
      --pid uint32        The PID of the process to display. This option should be used with [--ps-args | --ps-env-vars | --ps-memory-pages]
      --ps-args           Display the arguments for the specified pid
      --ps-env-vars       Display environment variables for the specified pid
      --ps-memory-pages   Display memory pages for the specified pid in hexdump format
      --stats             Print checkpointing statistics if available

```

Example of ps-args:

```bash
$ ./checkpointctl show --pid=1 --ps-args /path/to/checkpoint.tar.gz 

Displaying container checkpoint data from /path/to/checkpoint.tar.gz 

+---------------+---------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
|   CONTAINER   |              IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
+---------------+---------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
| awesome_booth | docker.io/library/ubuntu:latest | 695b77deb382 | crun    | 2023-03-08T08:45:33+03:00 | Podman | 2.8 MiB    | 309.0 KiB         |
+---------------+---------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+

Arguments of process 1 

/bin/bash 

```

Example of ps-env-vars:

```bash
$ ./checkpointctl show --pid=1 --ps-env-vars /path/to/checkpoint.tar.gz  

Displaying container checkpoint data from /path/to/checkpoint.tar.gz 

+---------------+---------------------------------+--------------+---------+---------------------------+--------+------------+-------------------+
|   CONTAINER   |              IMAGE              |      ID      | RUNTIME |          CREATED          | ENGINE | CHKPT SIZE | ROOT FS DIFF SIZE |
[...]

PR #95, opened 2023-07-29
feat: add `memparse` sub-command

behouba:
This PR introduces a new sub-command `memparse` ,  which allows analyzing processes memory pages. 
This new feature was discussed here #69. 

When used without any arguments, the command displays a table showing the memory sizes of processes. Here's an example:
```bash
$ checkpointctl memparse /path/to/checkpoints/jira.tar.gz 

Displaying processes memory sizes from /home/behouba/checkpoints/jira.tar.gz

+-----+--------------+-------------+
| PID | PROCESS NAME | MEMORY SIZE |
+-----+--------------+-------------+
|   1 | tini         | 100.0 KiB   |
+-----+--------------+-------------+
|   2 | java         | 553.5 MiB   |
+-----+--------------+-------------+
```

If a process ID (pid) is provided, the command prints the memory pages of that specific process in a hexdump-like format. For instance:

```bash
$ checkpointctl memparse /path/to/checkpoints/jira.tar.gz  --pid=1 | less

Displaying memory pages content for Process ID 1 from checkpoint: /home/behouba/checkpoints/jira.tar.gz

Address           Hexadecimal                                       ASCII            
-------------------------------------------------------------------------------------
00005633bb080000  f3 0f 1e fa 48 83 ec 08 48 8b 05 d1 4f 00 00 48  |....H...H...O..H|
00005633bb080010  85 c0 74 02 ff d0 48 83 c4 08 c3 00 00 00 00 00  |..t...H.........|
00005633bb080020  ff 35 b2 4e 00 00 f2 ff 25 b3 4e 00 00 0f 1f 00  |.5.N....%.N.....|
00005633bb080030  f3 0f 1e fa 68 00 00 00 00 f2 e9 e1 ff ff ff 90  |....h...........|
*
00005633bb0800a0  f3 0f 1e fa 68 07 00 00 00 f2 e9 71 ff ff ff 90  |....h......q....|
00005633bb0800b0  f3 0f 1e fa 68 08 00 00 00 f2 e9 61 ff ff ff 90  |....h......a....|
00005633bb0800c0  f3 0f 1e fa 68 09 00 00 00 f2 e9 51 ff ff ff 90  |....h......Q....|
00005633bb0800d0  f3 0f 1e fa 68 0a 00 00 00 f2 e9 41 ff ff ff 90  |....h......A....|
00005633bb0800e0  f3 0f 1e fa 68 0b 00 00 00 f2 e9 31 ff ff ff 90  |....h......1....|
````
This output can be written to a file instead of stdout using the `--output` flag.

@rst0git, @adrianreber, could you please take a look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,8 @@ Details on how to create checkpoints with the help of [CRIU][criu] can be found 
 
 ## Usage
 
+### `show` sub-command
+
 To display an overview of a checkpoint archive you can just use
 `checkpointctl show`:
 
@@ -41,6 +43,8 @@ $ checkpointctl show /var/lib/kubelet/checkpoints/checkpoint-counters_default-co
 +-----------+------------------------------------+--------------+---------+--------------------------------+--------+------------+------------+
 ```
 
+### `inspect` sub-command
+
 To retrieve low-level information about a container checkpoint, use the `checkpointctl inspect` command:
 
 ```console
@@ -63,6 +67,91 @@ awesome_booth
 ```
 
 For a complete list of flags supported, use `checkpointctl inspect --help`.
+
+### `memparse` sub-command
+
+To perform memory analysis of container checkpoints, you can use the `checkpointctl memparse` command.
+
+```console
+$ checkpointctl memparse /tmp/jira.tar.gz  --pid=1 | less
+
+Displaying memory pages content for Process ID 1 from checkpoint: /tmp/jira.tar.gz
+
+Address           Hexadecimal                                       ASCII            
+-------------------------------------------------------------------------------------
+00005633bb080000  f3 0f 1e fa 48 83 ec 08 48 8b 05 d1 4f 00 00 48  |....H...H...O..H|
+00005633bb080010  85 c0 74 02 ff d0 48 83 c4 08 c3 00 00 00 00 00  |..t...H.........|
+00005633bb080020  ff 35 b2 4e 00 00 f2 ff 25 b3 4e 00 00 0f 1f 00  |.5.N....%.N.....|
+00005633bb080030  f3 0f 1e fa 68 00 00 00 00 f2 e9 e1 ff ff ff 90  |....h...........|
+*
+00005633bb0800a0  f3 0f 1e fa 68 07 00 00 00 f2 e9 71 ff ff ff 90  |....h......q....|
+00005633bb0800b0  f3 0f 1e fa 68 08 00 00 00 f2 e9 61 ff ff ff 90  |....h......a....|
+00005633bb0800c0  f3 0f 1e fa 68 09 00 00 00 f2 e9 51 ff ff ff 90  |....h......Q....|
+00005633bb0800d0  f3 0f 1e fa 68 0a 00 00 00 f2 e9 41 ff ff ff 90  |....h......A....|
+00005633bb0800e0  f3 0f 1e fa 68 0b 00 00 00 f2 e9 31 ff ff ff 90  |....h......1....|
+```
+
+Here's an example of memory analysis of a PostgreSQL container. In this case, we start a PostgreSQL container with a password set to 'mysecret'. Then, we create a checkpoint of the container and use the `memparse` to find the stored password.
+
+```console
+$ sudo podman run --name postgres -e POSTGRES_PASSWORD=mysecret -d postgres
+$ sudo podman container checkpoint -l --export=/tmp/postgres.tar.gz
+$ sudo checkpointctl memparse --pid 1 /tmp/postgres.tar.gz | grep -B 1 -A 1 mysecret
+000055dd725c1e60  50 4f 53 54 47 52 45 53 5f 50 41 53 53 57 4f 52  |POSTGRES_PASSWOR|
+000055dd725c1e70  44 3d 6d 79 73 65 63 72 65 74 00 00 00 00 00 00  |D=mysecret......|
+000055dd725c1e80  00 00 00 00 00 00 00 00 31 00 00 00 00 00 00 00  |........1.......|
+```
+
+Here's another scenario, of memory analysis for a web application container. We start a vulnerable web application container, perform an arbitrary code execution attack, create a checkpoint for forensic analysis while leaving the container running, and finally analyze the checkpoint memory to identify the injected code.
+
+```console
+# Start vulnerable web application
+$ sudo podman run --name dsvw -p 1234:8000 -d quay.io/rst0git/dsvw
+
+# Perform arbitrary code execution attack: $(echo secret)
+$ curl "http://localhost:1234/?domain=www.google.com%3B%20echo%20secret"
+nslookup: can't resolve '(null)': Name does not resolve
+
+Name:      www.google.com
+Address 1: 142.250.187.228 lhr25s34-in-f4.1e100.net
+Address 2: 2a00:1450:4009:820::2004 lhr25s34-in-x04.1e100.net
+secret
+(reverse-i-search)`': ^C
+
+# Create a checkpoint for forensic analysis and leave the container running
+$ sudo podman container checkpoint --leave-running -l -e /tmp/dsvw.tar
+
+# Analyse checkpoint memory to identify the attacker's injected code
+$ sudo checkpointctl memparse --pid 1 /tmp/dsvw.tar | grep 'echo secret'
+00007faac5711f60  6f 6d 3b 20 65 63 68 6f 20 73 65 63 72 65 74 00  |om; echo secret.|
+```
+
+For larger processes, it's recommended to write the contents of process memory pages to a file rather than standard output.
+
+To get an overview of process memory sizes within the checkpoint, run `checkpointctl memparse` without arguments.
+
+```console
+$ sudo checkpointctl memparse /tmp/jira.tar.gz
+
+Displaying processes memory sizes from /tmp/jira.tar.gz
+
++-----+--------------+-------------+
+| PID | PROCESS NAME | MEMORY SIZE |
++-----+--------------+-------------+
+|   1 | tini         | 100.0 KiB   |
++-----+--------------+-------------+
+|   2 | java         | 553.5 MiB   |
++-----+--------------+-------------+
+```
+
+In this example, given the large size of the java process, it is better to write its output to a file.
+
+```console
+$ sudo checkpointctl memparse --pid=2 /tmp/jira.tar.gz --output=/tmp/java-memory-pages.txt
+Writing memory pages content for process ID 2 from checkpoint: /tmp/jira.tar.gz to file: /tmp/java-memory-pages.txt...
+```
+
+Please note that writing large memory pages to a file can take several minutes.
 
 ## Installing from source code
 
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -12,17 +12,18 @@ import (
 )
 
 var (
-	name      string
-	version   string
-	format    string
-	stats     bool
-	mounts    bool
-	pID       uint32
-	psTree    bool
-	psTreeCmd bool
-	psTreeEnv bool
-	files     bool
-	showAll   bool
+	name           string
+	version        string
+	format         string
+	stats          bool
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: checkpointctl.go [hunk 1]
```diff
@@ -262,3 +266,54 @@ func cleanupTasks(tasks []task) {
...
+	}
+
+	flags := cmd.Flags()
+
+	flags.Uint32VarP(
+		&pID,
+		"pid",
+		"p",
+		0,
+		"Specify the pid of a process to analyze",
```
snprajwal, line 285:
```suggestion
		"Specify the PID of a process to analyze",
```

checkpointctl.go
```diff
@@ -262,3 +266,54 @@ func cleanupTasks(tasks []task) {
...
+	)
+	flags.StringVarP(
+		&outputFilePath,
+		"output",
+		"o",
+		"",
+		"Specify a file where the ouput should be written",
```
snprajwal, line 292:
```suggestion
		"Specify the output file to be written to",
```

memparse.go
```diff
@@ -0,0 +1,181 @@
...
+
+	if outputFilePath != "" {
+		// Write to ouput to file if --output is provided
+		output, err = os.Create(outputFilePath)
+		if err != nil {
+			return err
+		}
+	} else {
+		compact = true // Use a compact format when writing the output to stdout
+		fmt.Printf("\nDisplaying memory pages content for Process ID %d from checkpoint: %s\n\n", pID, task.checkpointFilePath)
```
snprajwal, line 114:
```suggestion
		fmt.Printf("\nDisplaying memory pages content for process ID %d from checkpoint: %s\n\n", pID, task.checkpointFilePath)
```

memparse.go
```diff
@@ -0,0 +1,181 @@
...
+		if compact {
+			if prevAscii == ascii {
+				if duplicate == 1 {
+					fmt.Fprint(out, "*\n")
+				}
+				duplicate++
+			} else {
+				printHexdumpRow(out, vaddr, hex, ascii)
+				duplicate = 0
+			}
```
snprajwal, line 153: Instead of using an int to  track the duplicate, we can use a boolean `showDuplicate` instead
```suggestion
			if prevAscii == ascii {
				if showDuplicate {
					fmt.Fprint(out, "*\n")
					showDuplicate = false
				}
			} else {
				printHexdumpRow(out, vaddr, hex, ascii)
				showDuplicate = true
			}
```

memparse.go
```diff
@@ -0,0 +1,181 @@
...
+		} else {
+			printHexdumpRow(out, vaddr, hex, ascii)
+		}
+
+		vaddr += chunkSize
+		prevAscii = ascii
+	}
+}
+
+func printHexdumpRow(out io.Writer, vaddr uint64, hex string, ascii string) {
```
snprajwal, line 163: :thinking: this function is just a single line, does it need to be a separate function? The compiler will anyway inline it while optimising, but it still requires a jump from one location to another while reading the code.

memparse.go
```diff
@@ -0,0 +1,181 @@
...
+	fmt.Fprintf(out, "%016x  %s |%s|\n", vaddr, hex, ascii)
+}
+
+// generateHexAndAscii takes a byte slice and generates its hexadecimal and ASCII representations.
+func generateHexAndAscii(data []byte) (hex, ascii string) {
+	s := string(data)
```
snprajwal, line 169: This cast is not required, since we directly operate on the ASCII values of the characters, which is provided by the byte slice. The function will work identically if `data` is used in place of `s`.

memparse.go
```diff
@@ -0,0 +1,181 @@
...
+		if s[i] < 32 || s[i] >= 127 {
+			ascii += "."
+			hex += fmt.Sprintf("%02x ", s[i])
+		} else {
+			ascii += string(s[i])
+			hex += fmt.Sprintf("%02x ", s[i])
+		}
+	}
+
+	return hex, ascii
```
snprajwal, line 180: If we're using named return values in the function definition (L168), a `return` should suffice. While named return values don't reduce the code readability for something this small, it is still slightly unintuitive, and I tend to prefer using `(string, string)` in the function signature and explicitly initialising the return values at the beginning of the function body instead. Just a preference :)

memparse.go
behouba: I followed this approach to ensure consistent formatting for each row. If creating a new function would cause compiler overhead, I agree that it might be better not to declare a new function.
Thanks @snprajwal

checkpointctl.go
```diff
@@ -293,5 +311,9 @@ func memparse(cmd *cobra.Command, args []string) error {
 	}
 	defer cleanupTasks(tasks)
 
+	if pID != 0 {
```
rst0git, line 314: Should we extract only the relevant `pagemap-{PID}.img`, `pages-{PID}` and `mm-{PID}.img` files from the checkpoint archive when `--pid` has been specified? In addition, we could show appropriate error message if the specified PID does not exist in the checkpoint.

rst0git:
Great work @behouba!
It might be useful to extend `memparse` with support for [shared memory](https://criu.org/Shared_memory) in a subsequent pull request.

checkpointctl.go
behouba: Yes, I was thinking about this also. To do this, I need to make a change in `crit` (add a getter function to get the value of `pageNo` ).

rst0git: > To do this, I need to make a change in crit (add a getter function to get the value of pageNo). Should i open a PR on go-criu now?

It might be worth implementing this because it can significantly improve the performance for large container checkpoints. Also, when the `--pid` option is not used we don't need to extract `pages-` and `mm-`.

test/checkpointctl.bats
```diff
@@ -492,3 +492,60 @@ function teardown() {
...
+	cp test-imgs/pstree.img \
+		test-imgs/core-*.img \
+		test-imgs/pagemap-*.img \
+		test-imgs/pages-*.img \
+		test-imgs/mm-*.img "$TEST_TMP_DIR1"/checkpoint
+	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/test.tar . )
+	checkpointctl memparse "$TEST_TMP_DIR2"/test.tar --pid=9999 
+	[ "$status" -eq 1 ]
+	[[ ${lines[0]} == *"no process with PID 9999"* ]]
+}
```
snprajwal, line 551:
```suggestion
}

```

memparse.go
```diff
@@ -0,0 +1,202 @@
...
+		output = f
+		fmt.Printf("\nWriting memory pages content for process ID %d from checkpoint: %s to file: %s...\n",
+			pID, task.checkpointFilePath, outputFilePath,
+		)
+	} else {
+		compact = true // Use a compact format when writing the output to stdout
+		fmt.Printf("\nDisplaying memory pages content for process ID %d from checkpoint: %s\n\n", pID, task.checkpointFilePath)
+	}
+
+	fmt.Fprintf(output, "Address           Hexadecimal                                       ASCII            \n")
```
snprajwal, line 142: It would be better to use `fmt.Fprintln` here, since we're not formatting anything. `fmt.Fprintf` is slower as it needs to check the string for format specifiers and process every argument. Ditto for the line below (143)

container.go
```diff
@@ -348,3 +348,17 @@ func getPsEnvVars(checkpointOutputDir string, pid uint32) (envVars []string, err
...
 }
+
+// findPsTreeByPID performs a short-circuiting depth-first search to find
+// a process with a given PID in a process tree.
+func findPsTreeByPID(root *crit.PsTree, targetPID uint32) *crit.PsTree {
+	if root.PID == targetPID {
+		return root
+	}
+	for _, child := range root.Children {
+		if foundProcess := findPsTreeByPID(child, targetPID); foundProcess != nil {
```
snprajwal, line 359: 🤔 I think it is safe to name this variable as just `process`, since the condition implies that it is the found process if it is not nil. `foundProcess` sounds like a boolean to me.

container.go
```diff
@@ -348,3 +348,17 @@ func getPsEnvVars(checkpointOutputDir string, pid uint32) (envVars []string, err
 
 	return
 }
+
+// findPsTreeByPID performs a short-circuiting depth-first search to find
+// a process with a given PID in a process tree.
+func findPsTreeByPID(root *crit.PsTree, targetPID uint32) *crit.PsTree {
```
snprajwal, line 354: Would it be more appropriate to name this `findPs`? . It is well known that a PID is the primary identifier for a process, and we are searching for a process rather than a process tree.

snprajwal: It might also make sense to provide this as a method on `*PsTree` through go-criu, so that it can be used like `tree.FindPs(<pid>)`.

behouba: > It might also make sense to provide this as a method on `*PsTree` through go-criu,

This make sense to me as well. Can I add this  as `PsTree` method inside `go-criu/crit`  @rst0git , @adrianreber  ?

rst0git: Indeed, depth-first search for PsTree has many possible use cases. It would make sense to be in go-criu.

tree.go
```diff
@@ -121,21 +121,7 @@ func addDumpStatsToTree(tree treeprint.Tree, dumpStats *stats_pb.DumpStatsEntry)
...
-			for _, child := range root.Children {
-				if ps := dfs(child); ps != nil {
-					return ps
-				}
-			}
-			return nil
-		}
-
-		ps := dfs(psTree)
+		ps := psTree.FindPs(pID)
```
rst0git, line 124: Should we move this change in a separate pull request that also updates `go-criu` when https://github.com/checkpoint-restore/go-criu/pull/138 has been merged?

snprajwal (approved):
LGTM

rst0git:
@behouba Would it be possible to update go-criu in a separate pull request?

rst0git:
@behouba Could you add a section about memory analysis to the README file?
It might be good to describe how users can use `checkpointctl memparse` with an example:
```console
$ sudo podman run --name postgres -e POSTGRES_PASSWORD=mysecret -d postgres
$ sudo podman container checkpoint -l --export=/tmp/postgres.tar.gz
$ sudo sudo ./checkpointctl memparse --pid 1 /tmp/postgres.tar.gz | grep mysecret
000055f9deed8e70  44 3d 6d 79 73 65 63 72 65 74 00 00 00 00 00 00  |D=mysecret......|
```

behouba:
> @behouba Could you add a section about memory analysis to the README file?

@rst0git PTAL, and thank you for providing such good examples.

README.md
```diff
@@ -64,6 +68,91 @@ awesome_booth
...
+$ sudo checkpointctl memparse --pid 1 /tmp/dsvw.tar | grep 'echo secret'
+00007faac5711f60  6f 6d 3b 20 65 63 68 6f 20 73 65 63 72 65 74 00  |om; echo secret.|
+```
+
+For larger processes, it's recommended to write the contents of process memory pages to a file rather than standard output.
+
+To get an overview of process memory sizes within the checkpoint, run `checkpointctl memparse` without arguments.
+
+```console
+$ sudo ./checkpointctl memparse /tmp/jira.tar.gz
```
rst0git, line 134:
```Suggestion
$ sudo checkpointctl memparse /tmp/jira.tar.gz
```

README.md
```diff
@@ -64,6 +68,91 @@ awesome_booth
...
+|   1 | tini         | 100.0 KiB   |
++-----+--------------+-------------+
+|   2 | java         | 553.5 MiB   |
++-----+--------------+-------------+
+```
+
+In this example, given the large size of the java process, it is better to write it output to a file.
+
+```console
+$ sudo ./checkpointctl memparse --pid=2 /tmp/jira.tar.gz --output=/tmp/java-memory-pages.txt
```
rst0git, line 150:
```Suggestion
$ sudo checkpointctl memparse --pid=2 /tmp/jira.tar.gz --output=/tmp/java-memory-pages.txt
```

rst0git (approved):
LGTM

PR #96, opened 2023-08-04
inspect: enable --ps-tree-{env,cmd} with --all

rst0git:
This patch enables `--ps-tree-cmd` and `--ps-tree-env` options when the `inspect` command is used with `--all`. This change provides more complete information about the processes included in a container checkpoint.


rst0git:
@behouba @snprajwal PTAL

behouba (approved):
LGTM

PR #98, opened 2023-08-07
stats: use time units in human-friendly format

rst0git:
This pull request transforms the CRIU statistics into human-readable format.

Example:
```
httpd
├── Image: docker.io/library/httpd:2.4
├── ID: 04b6be846ef4535bf9027ae36b7f3efeda78f2c314c4d9694b09da70db042b88
├── Runtime: crun
├── Created: 2023-08-07T13:02:38+01:00
├── Engine: Podman
├── Checkpoint size: 5.2 MiB
├── Root FS diff size: 2.0 KiB
└── CRIU dump statistics
    ├── Freezing time: 3.781 ms
    ├── Frozen time: 116.54 ms
    ├── Memdump time: 22.666 ms
    ├── Memwrite time: 3.382 ms
    ├── Pages scanned: 1498870
    └── Pages written: 1213
```

### DIFF
--- a/utils.go
+++ b/utils.go
@@ -1,0 +1,30 @@
+package main
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+func FormatTime(microseconds uint32) string {
+	if microseconds < 1000 {
+		return fmt.Sprintf("%d µs", microseconds)
+	}
+
+	var value float64
+	var unit string
+
+	if microseconds < 1000000 {
+		value = float64(microseconds) / 1000
+		unit = "ms"
+	} else {
+		duration := time.Duration(microseconds) * time.Microsecond
+		value = duration.Seconds()
+		unit = "s"
+	}
+
+	// Trim trailing zeros and dot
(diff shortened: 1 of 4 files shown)


utils.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	duration := time.Duration(microseconds) * time.Microsecond
+
+	var value float64
+	var unit string
+
+	switch {
+	case microseconds < 1000000:
+		value = float64(microseconds) / 1000
```
snprajwal, line 21: Do you think we can instead use an approach similar to what [`countBytes`](https://github.com/checkpoint-restore/go-criu/blob/master/crit/utils.go#L46) does in go-criu?

rst0git: `countBytes()` is good but this function is simpler — it converts only `µs` to `ms` or `s`. Thus, we have fewer conditions and adding a for loop would make it more complicated.

snprajwal (approved):
I never noticed that it's not in a human friendly format °_° nice catch! LGTM

PR #94, opened 2023-07-28
feat: display sockets in process tree

snprajwal:
This is now supported with the `--sockets` flag.

PR #102, opened 2023-08-14
tree: use sub-nodes for open files and env vars

rst0git:
When using inspect with `--all`, the information about open files and environment variables appears in format that may not be easy for users to interpret and understand. This pull request adds appropriate labels to address this issue.

Example:

```
├── Process tree
│   └── [1]  /bin/sh -c i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done
│       ├── Environment variables
│       │   ├── container=podman
│       │   ├── PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
│       │   ├── TERM=xterm
│       │   ├── HOME=/root
│       │   └── HOSTNAME=a8e8196b24f8
│       └── Open files
│           ├── [REG 0]  /dev/null
│           ├── [PIPE 1]  pipe[1654003]
│           ├── [PIPE 2]  pipe[1654004]
│           ├── [cwd]  /
│           └── [root]  /
```

### DIFF
--- a/test/checkpointctl.bats
+++ b/test/checkpointctl.bats
@@ -329,7 +329,7 @@ function teardown() {
 	[ "$status" -eq 0 ]
 	[[ ${lines[9]} == *"Process tree"* ]]
 	[[ ${lines[10]} == *"piggie"* ]]
-	[[ ${lines[11]} == *"="* ]]
+	[[ ${lines[12]} == *"="* ]]
 }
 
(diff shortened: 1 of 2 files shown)


rst0git:
@behouba @snprajwal PTAL

snprajwal (approved):
Nice upgrade! I feel this is a good idea to do even with sockets in #94, I'll update it there.

behouba (approved):
LGTM

Issue #77, opened 2023-06-21
Extend `inspect` output format with JSON


behouba:
Currently `checkpointctl inspect` only support the tree format. It might be useful to also be able to render the output in a JSON format (`--format=json`). This would enable easier integration with other tools to further process the output of checkpointctl.

Suggested by: @rst0git

snprajwal:
This would be trivial to do for the existing features coming from CRIT (process tree, file descriptors, etc) since the original format is JSON itself. At most, we will have to use a custom struct to remove some of the fields that are not relevant.

Parthiba-Hazra:
Hi @behouba @snprajwal @rst0git 
if the issue is not been worked yet, can I take it ?

rst0git:
@Parthiba-Hazra Sure, please feel free to open a pull request!

PR #113, opened 2023-12-26
Extend inspect output format with json

Parthiba-Hazra:
issue: #77

### DIFF
--- a/json.go
+++ b/json.go
@@ -50,9 +50,28 @@ type SkNode struct {
 	OpenSockets []SocketNode `json:"open_sockets,omitempty"`
 }
 
+type SkData struct {
+	Type       string `json:"type,omitempty"`
+	Address    string `json:"address,omitempty"`
+	State      string `json:"state,omitempty"`
+	Source     string `json:"src,omitempty"`
+	SourcePort uint32 `json:"src_port,omitempty"`
+	Dest       string `json:"dst,omitempty"`
+	DestPort   uint32 `json:"dst_port,omitempty"`
+	SendBuf    string `json:"send_buf,omitempty"`
+	RecvBuf    string `json:"recv_buf,omitempty"`
+}
+
 type SocketNode struct {
-	Protocol string `json:"protocol"`
-	Data     string `json:"data"`
+	Protocol string `json:"protocol,omitempty"`
+	Data     SkData `json:"data,omitempty"`
+}
+
+var socketDataFuncs = map[string]func(*crit.Socket) SkData{
+	"UNIXSK":    getUnixSkData,
+	"INETSK":    getInetSkData,
+	"PACKETSK":  getPacketSkData,
+	"NETLINKSK": getNetlinkSkData,
 }
 
 type DisplayNode struct {
@@ -168,7 +187,10 @@ func renderJSONView(tasks []task) error {
 				return fmt.Errorf("failed to get sockets: %w", err)
 			}
 
-			node.Sockets = buildJSONSks(fds)
+			node.Sockets, err = buildJSONSks(fds)
+			if err != nil {
+				return fmt.Errorf("failed to build sockets: %w", err)
+			}
 		}
 
 		if mounts {
@@ -302,7 +324,7 @@ func buildJSONFds(fds []*crit.Fd) []FdNode {
 	return result
 }
 
-func buildJSONSks(sks []*crit.Sk) []SkNode {
+func buildJSONSks(sks []*crit.Sk) ([]SkNode, error) {
 	var result []SkNode
 
 	for _, sk := range sks {
@@ -312,11 +334,14 @@ func buildJSONSks(sks []*crit.Sk) []SkNode {
 
 		var sockets []SocketNode
 		for _, socket := range sk.Sockets {
-			socketNode := SocketNode{
-				Protocol: socket.Protocol,
-				Data:     getDataForSocket(socket),
+			socketData, err := getDataForSocket(socket)
+			if err != nil {
+				return nil, fmt.Errorf("error getting data for socket: %w", err)
 			}
-			sockets = append(sockets, socketNode)
+			sockets = append(sockets, SocketNode{
+				Protocol: socket.Protocol,
+				Data:     socketData,
+			})
 		}
 
 		if len(sockets) > 0 {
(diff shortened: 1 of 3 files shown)


json.go
```diff
@@ -0,0 +1,266 @@
...
+	return map[string]interface{}{"root": rootNode}
+}
+
+func buildJSONPsNode(psTree *crit.PsTree, checkpointOutputDir string) (map[string]interface{}, error) {
+	node := map[string]interface{}{
+		"pid":  psTree.PID,
+		"comm": psTree.Comm,
+	}
+
+	cmdline, err := getCmdline(checkpointOutputDir, psTree.PID)
```
behouba, line 145: When adding the `cmdline` key, it may be good to first check if the `ps-tree-cmd` flag is set to true. Without this check, the current implementation will fail when using only `--ps-tree` flag since memory pages files won't be extracted in this case.

json.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+
+		node := map[string]interface{}{
+			"container_name": info.containerInfo.Name,
+			"image":          info.configDump.RootfsImageName,
+			"id":             info.configDump.ID,
+			"runtime":        info.configDump.OCIRuntime,
+			"created":        info.containerInfo.Created,
+			"engine":         info.containerInfo.Engine,
+			"ip":             info.containerInfo.IP,
```
behouba, line 30: It may be good to set `ip` and `mac` keys only when there is a value.

json.go
```diff
@@ -0,0 +1,266 @@
...
+			"source":      data.Source,
+		}
+		result = append(result, mountNode)
+	}
+
+	return result
+}
+
+func buildJSONPsTree(psTree *crit.PsTree, checkpointOutputDir string) map[string]interface{} {
+	rootNode, err := buildJSONPsNode(psTree, checkpointOutputDir)
```
behouba, line 131: We should also check the `--pid` flag somewhere here to ensure that the process tree to be displayed starts from the specified PID.

behouba (commented):
Hi @Parthiba-Hazra, thank you for taking the time to work on this issue. I have added some comments.

json.go
Parthiba-Hazra: got it, thank you for the quick review.

snprajwal (commented):
This looks good for the most part, but I'm not too convinced about using a `map[string]interface{}` for storing the JSON object. `interface{}` adds a runtime cost to ensure type safety, and might end up having significant overhead for bigger checkpoints. I think marshaling the data in chunks into a `json.RawMessage` object and doing some string slicing will be considerably more efficient. There's something similar implemented in <https://github.com/checkpoint-restore/go-criu/blob/master/crit/image.go#L85-L97>, maybe you can check it out and see if that can be used here.

Parthiba-Hazra:
> This looks good for the most part, but I'm not too convinced about using a `map[string]interface{}` for storing the JSON object. `interface{}` adds a runtime cost to ensure type safety, and might end up having significant overhead for bigger checkpoints. I think marshaling the data in chunks into a `json.RawMessage` object and doing some string slicing will be considerably more efficient. There's something similar implemented in https://github.com/checkpoint-restore/go-criu/blob/master/crit/image.go#L85-L97, maybe you can check it out and see if that can be used here.

yes you are right, should I go for custom struct instead of interfaces?

snprajwal:
> yes you are right, should I go for custom struct instead of interfaces?

Yes, that would be ideal!

behouba (commented):
@Parthiba-Hazra, let's not add fix-up commits. Instead, amend the initial commit or add new self contained commits. This will ensure that that the commit history is clean. Checkout the [CRIU contribution guidelines](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md) for more details.
It's also a good idea to test the JSON format.

Parthiba-Hazra:
> @Parthiba-Hazra, let's not add fix-up commits. Instead, amend the initial commit or make add new self contained commits. This will ensure that that the commit history is clean. Checkout the [CRIU contribution guidelines](https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md) for more details. It's also a good idea to test the JSON format.

sorry if I am wrong but I didn't perform fixup commits, I just added new commits on top of other commits.

adrianreber:
The second commit looks indeed like a fixup. You introduce a new feature in the first commit and then you say "minor changes" in the second commit. If you introduce a new feature in one PR you can separate logical changes in different commits but fixups like "minor changes" should be part of the first commit. No need to create a second commit. You can include the fixup in the original commit as long as it is a PR and not yet merged. Once it is merged and it needs to change the situation is different. A commit message like "minor changes" would probably also be questionable and would require a better description of the change.

Parthiba-Hazra:
> The second commit looks indeed like a fixup. You introduce a new feature in the first commit and then you say "minor changes" in the second commit. If you introduce a new feature in one PR you can separate logical changes in different commits but fixups like "minor changes" should be part of the first commit. No need to create a second commit. You can include the fixup in the original commit as long as it is a PR and not yet merged. Once it is merged and it needs to change the situation is different. A commit message like "minor changes" would probably also be questionable and would require a better description of the change.

yea got it, sorry for the inconvenience, I just got to know about this fixup commit thing, will keep in my mind for sure. should I create another pr with proper commits?

snprajwal:
> should I create another pr with proper commits?

Not necessary, you can rebase your branch to fix this. In your case, you need to squash your two recent commits into the oldest commit (extend inspect output format with json). You can do this with the below steps:
- `git rebase -i HEAD~3`
- Change the latest two commits from `pick` to `s` or `squash` in the rebase window
- Ensure the new commit contains the correct message (by default, git combines the messages of the `pick` and `squash` commits)
- `git push --force-with-lease`

Parthiba-Hazra:
> > should I create another pr with proper commits?
> 
> Not necessary, you can rebase your branch to fix this. In your case, you need to squash your two recent commits into the oldest commit (extend inspect output format with json). You can do this with the below steps:
> 
> * `git rebase -i HEAD~3`
> * Change the latest two commits from `pick` to `s` or `squash` in the rebase window
> * Ensure the new commit contains the correct message (by default, git combines the messages of the `pick` and `squash` commits)
> * `git push --force-with-lease`

thanks prajwal for the guidance here, really forgot about squash.

json.go
```diff
@@ -0,0 +1,357 @@
...
+			dumpStats, err := crit.GetDumpStats(task.outputDir)
+			if err != nil {
+				return fmt.Errorf("failed to get dump statistics: %w", err)
+			}
+
+			statsNode := StatsNode{
+				FreezingTime: FormatTime(dumpStats.GetFreezingTime()),
+				FrozenTime:   FormatTime(dumpStats.GetFrozenTime()),
+				MemdumpTime:  FormatTime(dumpStats.GetMemdumpTime()),
+				MemwriteTime: FormatTime(dumpStats.GetMemwriteTime()),
```
rst0git, line 135: same as above - in JSON format would be better to use the original integer value and avoid `FormatTime()`

rst0git (commented):
@Parthiba-Hazra nice work!

json.go
```diff
@@ -0,0 +1,367 @@
...
+}
+
+type SkNode struct {
+	PID         uint32       `json:"pid"`
+	OpenSockets []SocketNode `json:"open_sockets,omitempty"`
+}
+
+type SocketNode struct {
+	Protocol string `json:"protocol"`
+	Data     string `json:"data"`
```
rst0git, line 55: It might be good to also define a type socket data.

Example:
```Go
type SkData struct {
	Type       string `json:"type,omitempty"`
	Address    string `json:"address,omitempty"`
	State      string `json:"state,omitempty"`
	Source     string `json:"src,omitempty"`
	SourcePort uint32 `json:"src_port,omitempty"`
	Dest       string `json:"dst,omitempty"`
	DestPort   uint32 `json:"dst_port,omitempty"`
	SendBuf    string `json:"send_buf,omitempty"`
	RecvBuf    string `json:"recv_buf,omitempty"`
}

type SocketNode struct {
	Protocol string `json:"protocol,omitempty"`
	Data     SkData `json:"data,omitempty"`
}

var socketDataFuncs = map[string]func(*crit.Socket) (SkData, error){
	"UNIXSK":    getUnixSkData,
	"INETSK":    getInetSkData,
	"PACKETSK":  getPacketSkData,
	"NETLINKSK": getNetlinkSkData,
}

...

func buildJSONSks(sks []*crit.Sk) ([]SkNode, error) {
	var result []SkNode

	for _, sk := range sks {
		skNode := SkNode{
			PID: sk.PId,
		}

		var sockets []SocketNode
		for _, socket := range sk.Sockets {
			socketData, err := getDataForSocket(socket)
			if err != nil {
				return nil, fmt.Errorf("error getting data for socket: %w", err)
			}
			sockets = append(sockets, SocketNode{
				Protocol: socket.Protocol,
				Data:     socketData,
			})
		}

		if len(sockets) > 0 {
			skNode.OpenSockets = sockets
		}

		result = append(result, skNode)
	}

	return result, nil
}

func getDataForSocket(socket *crit.Socket) (SkData, error) {
	dataFunc, found := socketDataFuncs[socket.FdType]
	if !found {
		return SkData{}, fmt.Errorf("unsupported socket type: %s", socket.FdType)
	}
	return dataFunc(socket)
}

func getUnixSkData(socket *crit.Socket) (SkData, error) {
	return SkData{
		Type:    "UNIX",
		Address: socket.SrcAddr,
	}, nil
}

func getInetSkData(socket *crit.Socket) (SkData, error) {
	return SkData{
		Type:       socket.Protocol,
		State:      socket.State,
		Source:     socket.SrcAddr,
		SourcePort: socket.SrcPort,
		Dest:       socket.DestAddr,
		DestPort:   socket.DestPort,
		SendBuf:    socket.SendBuf,
		RecvBuf:    socket.RecvBuf,
	}, nil
}

func getPacketSkData(socket *crit.Socket) (SkData, error) {
	return SkData{
		SendBuf: socket.SendBuf,
		RecvBuf: socket.RecvBuf,
	}, nil
}

func getNetlinkSkData(socket *crit.Socket) (SkData, error) {
	return SkData{
		SendBuf: socket.SendBuf,
		RecvBuf: socket.RecvBuf,
	}, nil
}
```

Parthiba-Hazra: yeah sure,

adrianreber:
@Parthiba-Hazra Please take a look at the failed CI results. Also take a look at adding tests. Currently codecov says 0% patch test coverage.

rst0git (approved):

PR #114, opened 2023-12-28
tree: skip empty env variables branch

rst0git:
In some cases, processes may remove all environment variables inherited from the parent process (e.g., [nginx](https://nginx.org/en/docs/ngx_core_module.html#env)). This pull request adds a conditional statement to show the environment variables tree branch only when necessary.


snprajwal (approved):

behouba (approved):
LGTM

Issue #54, opened 2023-05-17
Add support for list command


rst0git:
There are several standard locations where container engines like Podman and CRI-O store container checkpoints. However, these locations are not well documented and it might be difficult for users to find them. It would be very useful to have a command like `checkpointctl list` that would be able to list container checkpoints created by Podman and CRI-O.

Nitishupkr:
i take up this issue

rst0git:
Hi @Nitishupkr, you are more than welcome to work on this GitHub issue. Please let us know if you have any questions.

Parthiba-Hazra:
@Nitishupkr you still working on this? if not then I can take on.

Parthiba-Hazra:
@rst0git if I am not wrong then container engine store container checkpoints here - 
podman - /var/lib/containers/storage/overlay-containers
CRI-O - /var/run/crio/checkpoints

for the list command are we gonna print the full path of checkpoint containers?
and is there any other directories we also need to check? and what about the checkpoints that are store in the directory specified by the users during the container checkpoint command execution?

rst0git:
@Parthiba-Hazra As a starting point, we can add `checkpointctl list` command that shows checkpoints stored in `/var/lib/kubelet/checkpoints/`. This path is used by [Kubernetes/CRI-O](https://kubernetes.io/blog/2022/12/05/forensic-container-checkpointing-alpha/) and contains files with the following naming convention:

```
checkpoint-<pod-name>_<namespace-name>-<container-name>-<timestamp>.tar
```

Example:
```
checkpoint-counter-tl58j_default-counter-2023-10-12T11:02:56+01:00.tar
```

> for the list command are we gonna print the full path of checkpoint containers?

It would good to use file name parsing and output a table similar to the `checkpointctl show` command. This functionality is particularly useful in scenarios where automated periodic checkpointing is used as a fault-tolerance mechanism, leading to many container checkpoints created in this directory. 

> what about the checkpoints that are store in the directory specified by the users during the container checkpoint command execution?

Once the list command has been merged, we could extend checkpointctl with support for configuration files that enable users to change the default path or to specify additional locations.

@adrianreber Do you have any thoughts?

Parthiba-Hazra:
[...]

PR #115, opened 2024-01-03
Add support for list command to get the list of container checkpoints

Parthiba-Hazra:
issue: #54

### DIFF
--- a/cmd/list.go
+++ b/cmd/list.go
@@ -1,0 +1,101 @@
+// SPDX-License-Identifier: Apache-2.0
+
+// This file is used to show the list of container checkpoints
+
+package cmd
+
+import (
+	"fmt"
+	"log"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/checkpoint-restore/checkpointctl/internal"
+	"github.com/olekukonko/tablewriter"
+	"github.com/spf13/cobra"
+)
+
+var (
+	defaultCheckpointPath     = "/var/lib/kubelet/checkpoints/"
+	additionalCheckpointPaths []string
+)
+
+func List() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list",
+		Short: "List checkpoints stored in the default and additional paths",
+		RunE:  list,
+	}
+
+	flags := cmd.Flags()
+	flags.StringSliceVarP(
+		&additionalCheckpointPaths,
+		"paths",
+		"p",
+		[]string{},
+		"Specify additional paths to include in checkpoint listing",
+	)
+
+	return cmd
+}
+
+func list(cmd *cobra.Command, args []string) error {
+	allPaths := append([]string{defaultCheckpointPath}, additionalCheckpointPaths...)
+	showTable := false
+
+	table := tablewriter.NewWriter(os.Stdout)
+	header := []string{
+		"Namespace",
+		"Pod",
+		"Container",
+		"Engine",
+		"Time Checkpointed",
+		"Checkpoint Name",
+	}
+
+	table.SetHeader(header)
+	table.SetAutoMergeCells(false)
+	table.SetRowLine(true)
+
+	for _, checkpointPath := range allPaths {
+		files, err := filepath.Glob(filepath.Join(checkpointPath, "checkpoint-*"))
+		if err != nil {
+			return err
+		}
+
+		if len(files) == 0 {
+			continue
+		}
+
+		showTable = true
+		fmt.Printf("Listing checkpoints in path: %s\n", checkpointPath)
+
+		for _, file := range files {
+			chkptConfig, err := internal.ExtractConfigDump(file)
+			if err != nil {
+				log.Printf("Error extracting information from %s: %v\n", file, err)
+				continue
+			}
+
+			row := []string{
+				chkptConfig.Namespace,
+				chkptConfig.Pod,
+				chkptConfig.Container,
+				chkptConfig.ContainerManager,
+				chkptConfig.Timestamp.Format(time.RFC822),
+				filepath.Base(file),
+			}
+
+			table.Append(row)
+		}
+	}
+
+	if !showTable {
+		fmt.Println("No checkpoints found")
+		return nil
+	}
+
+	table.Render()
+	return nil
+}
--- a/docs/checkpointctl-list.adoc
+++ b/docs/checkpointctl-list.adoc
@@ -1,0 +1,26 @@
+= checkpointctl-list(1)
+include::footer.adoc[]
+
+== Name
+
+*checkpointctl-list* - List checkpoints stored in the default and additional paths
+
+== Synopsis
+
+*checkpointctl list* [_OPTION_]... _FOLDER_...
+
+== Options
+
+*-h*, *--help*::
+  Show help for checkpointctl list
+
+*-p*, *--path*::
+  Specify additional paths to include in checkpoint listing
+
+== Default Path
+
+The default path for checking checkpoints is `/var/lib/kubelet/checkpoints/`.
+
+== See also
+
+checkpointctl(1)
--- a/internal/config_extractor.go
+++ b/internal/config_extractor.go
@@ -1,0 +1,76 @@
+package internal
+
+import (
+	"encoding/json"
+	"log"
+	"os"
+	"path/filepath"
+	"time"
+
+	metadata "github.com/checkpoint-restore/checkpointctl/lib"
+)
+
+type ChkptConfig struct {
+	Namespace        string
+	Pod              string
+	Container        string
+	ContainerManager string
+	Timestamp        time.Time
+}
+
+func ExtractConfigDump(checkpointPath string) (*ChkptConfig, error) {
+	tempDir, err := os.MkdirTemp("", "extracted-checkpoint")
+	if err != nil {
+		return nil, err
+	}
+	defer os.RemoveAll(tempDir)
+
+	filesToExtract := []string{"spec.dump", "config.dump"}
+	if err := UntarFiles(checkpointPath, tempDir, filesToExtract); err != nil {
+		log.Printf("Error extracting files from archive %s: %v\n", checkpointPath, err)
+		return nil, err
+	}
+
+	specDumpPath := filepath.Join(tempDir, "spec.dump")
+	specContent, err := os.ReadFile(specDumpPath)
+	if err != nil {
+		log.Printf("Error reading spec.dump file: %v\n", err)
+		return nil, err
+	}
+
+	configDumpPath := filepath.Join(tempDir, "config.dump")
+	configContent, err := os.ReadFile(configDumpPath)
+	if err != nil {
+		log.Printf("Error reading config.dump file: %v\n", err)
+		return nil, err
+	}
+
+	return extractConfigDumpContent(configContent, specContent)
+}
+
+func extractConfigDumpContent(configContent []byte, specContent []byte) (*ChkptConfig, error) {
+	var spec metadata.Spec
+	var config metadata.ContainerConfig
+
+	if err := json.Unmarshal(configContent, &config); err != nil {
+		return nil, err
+	}
+
+	if err := json.Unmarshal(specContent, &spec); err != nil {
+		return nil, err
+	}
+
+	namespace := spec.Annotations["io.kubernetes.pod.namespace"]
+	timestamp := config.CheckpointedAt
+	pod := spec.Annotations["io.kubernetes.pod.name"]
+	container := spec.Annotations["io.kubernetes.container.name"]
+	containerManager := spec.Annotations["io.container.manager"]
+
+	return &ChkptConfig{
+		Namespace:        namespace,
+		Pod:              pod,
+		Container:        container,
+		ContainerManager: containerManager,
+		Timestamp:        timestamp,
+	}, nil
+}
--- a/internal/config_extractor_test.go
+++ b/internal/config_extractor_test.go
@@ -1,0 +1,33 @@
+package internal
+
+import (
+	"os"
+	"testing"
+	"time"
+)
+
+func TestExtractConfigDumpContent(t *testing.T) {
+	configContent, err := os.ReadFile("../test/data/list_config_spec.dump/config.dump")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	specContent, err := os.ReadFile("../test/data/list_config_spec.dump/spec.dump")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	chkptConfig, err := extractConfigDumpContent(configContent, specContent)
+	if err != nil {
+		t.Fatalf("ExtractConfigDumpContent failed: %v", err)
+	}
+
+	expectedNamespace := "default"
+	expectedPod := "pod-name"
+	expectedContainer := "container-name"
+	expectedContainerManager := "cri-o"
+	expectedTimestamp := time.Date(2024, 1, 28, 0, 10, 45, 673538606, time.FixedZone("", 19800))
+	if chkptConfig.Namespace != expectedNamespace || chkptConfig.Pod != expectedPod || chkptConfig.Container != expectedContainer || !chkptConfig.Timestamp.Equal(expectedTimestamp) || chkptConfig.ContainerManager != expectedContainerManager {
+		t.Errorf("ExtractConfigDumpContent returned unexpected values: namespace=%s, pod=%s, container=%s, timestamp=%v", chkptConfig.Namespace, chkptConfig.Pod, chkptConfig.Container, chkptConfig.Timestamp)
+	}
+}
--- a/test/checkpointctl.bats
+++ b/test/checkpointctl.bats
@@ -640,3 +640,60 @@ function teardown() {
 	run bash -c "$CHECKPOINTCTL inspect $TEST_TMP_DIR2/test.tar --format=json --sockets | test_socket_src_port"
 	[ "$status" -eq 0 ]
 }
+
+@test "Run checkpointctl list with empty directory" {
+    mkdir "$TEST_TMP_DIR1"/empty
+    checkpointctl list -p "$TEST_TMP_DIR1"/empty/
+    [ "$status" -eq 0 ]
+    [[ ${lines[0]} == *"No checkpoints found"* ]]
+}
+
+@test "Run checkpointctl list with non existing directory" {
+	checkpointctl list -p /does-not-exist
+	[ "$status" -eq 0 ]
+	[[ ${lines[0]} == *"No checkpoints found"* ]]
+}
+
+@test "Run checkpointctl list with empty tar file" {
+	touch "$TEST_TMP_DIR1"/checkpoint-nginx-empty.tar
+	checkpointctl list -p "$TEST_TMP_DIR1"
+	[ "$status" -eq 0 ]
+	[[ "${lines[1]}" == *"Error reading spec.dump file"* ]]
+	[[ "${lines[2]}" == *"Error extracting information"* ]]
+}
+
+@test "Run checkpointctl list with tar file with valid spec.dump and empty config.dump" {
+	touch "$TEST_TMP_DIR1"/config.dump
+	cp data/list_config_spec.dump/spec.dump "$TEST_TMP_DIR1"
+	mkdir "$TEST_TMP_DIR1"/checkpoint
+	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/checkpoint-config.tar . )
+	checkpointctl list -p "$TEST_TMP_DIR2"
+	[ "$status" -eq 0 ]
+	[[ "${lines[1]}" == *"Error extracting information from $TEST_TMP_DIR2/checkpoint-config.tar: unexpected end of JSON input"* ]]
+}
+
+@test "Run checkpointctl list with tar file with valid config.dump and empty spec.dump" {
+	touch "$TEST_TMP_DIR1"/spec.dump
+	cp data/list_config_spec.dump/config.dump "$TEST_TMP_DIR1"
+	mkdir "$TEST_TMP_DIR1"/checkpoint
+	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/checkpoint-config.tar . )
+	checkpointctl list -p "$TEST_TMP_DIR2"
+	[ "$status" -eq 0 ]
+	[[ ${lines[1]} == *"Error extracting information from $TEST_TMP_DIR2/checkpoint-config.tar: unexpected end of JSON input" ]]
+}
+
+@test "Run checkpointctl list with tar file with valid config.dump and spec.dump" {
+	cp data/list_config_spec.dump/config.dump "$TEST_TMP_DIR1"
+	cp data/list_config_spec.dump/spec.dump "$TEST_TMP_DIR1"
+	mkdir "$TEST_TMP_DIR1"/checkpoint
+	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/checkpoint-valid-config.tar . )
+	jq '.["annotations"]["io.kubernetes.pod.name"] = "modified-pod-name"' "$TEST_TMP_DIR1"/spec.dump > "$TEST_TMP_DIR1"/spec_modified.dump
+	mv "$TEST_TMP_DIR1"/spec_modified.dump "$TEST_TMP_DIR1"/spec.dump
+	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/checkpoint-valid-config-modified.tar . )
+	checkpointctl list -p "$TEST_TMP_DIR2"
+	[ "$status" -eq 0 ]
+	[[ "${lines[4]}" == *"| default   | modified-pod-name | container-name | cri-o  |"* ]]
+	[[ "${lines[4]}" == *"| checkpoint-valid-config-modified.tar |"* ]]
+	[[ "${lines[6]}" == *"| default   | pod-name          | container-name | cri-o  |"* ]]
+	[[ "${lines[6]}" == *"| checkpoint-valid-config.tar          |"* ]]
+}
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -29,6 +29,8 @@ func main() {
 
 	rootCommand.AddCommand(cmd.MemParse())
 
+	rootCommand.AddCommand(cmd.List())
+
 	rootCommand.Version = version
 
 	if err := rootCommand.Execute(); err != nil {
--- a/lib/metadata.go
+++ b/lib/metadata.go
@@ -48,6 +48,10 @@ type ContainerConfig struct {
 	Restored        bool      `json:"restored"`
 }
 
+type Spec struct {
+	Annotations map[string]string `json:"annotations,omitempty"`
+}
+
 type ContainerdStatus struct {
 	CreatedAt  int64
 	StartedAt  int64
--- a/test/data/list_config_spec.dump/config.dump
+++ b/test/data/list_config_spec.dump/config.dump
@@ -1,0 +1,12 @@
+{
+  "id": "6924be1bd90c23f10e2667102b0ee0f74f09bba78b6661871e733cb3b1737821",
+  "name": "k8s_container-name_deployment-name_default_6975ee47-6765-45dc-9a2b-1e38d51031f7_0",
+  "rootfsImage": "docker.io/library/nginx@sha256:161ef4b1bf7effb350a2a9625cb2b59f69d54ec6059a8a155a1438d0439c593c",
+  "rootfsImageRef": "a8758716bb6aa4d90071160d27028fe4eaee7ce8166221a97d30440c8eac2be6",
+  "rootfsImageName": "docker.io/library/nginx:latest",
+  "runtime": "runc",
+  "createdTime": "2024-01-27T14:45:26.083444055Z",
+  "checkpointedTime": "2024-01-28T00:10:45.673538606+05:30",
+  "restoredTime": "0001-01-01T00:00:00Z",
+  "restored": false
+}
--- a/test/data/list_config_spec.dump/spec.dump
+++ b/test/data/list_config_spec.dump/spec.dump
@@ -1,0 +1,9 @@
+{
+	"annotations": {
+	    "io.container.manager": "cri-o",
+		"io.kubernetes.container.hash": "1511917a",
+		"io.kubernetes.container.name": "container-name",
+		"io.kubernetes.pod.name": "pod-name",
+		"io.kubernetes.pod.namespace": "default"
+	}
+}


adrianreber:
Please make sure you have tests for the functionality you added.

Parthiba-Hazra:
> Please make sure you have tests for the functionality you added.

yes sure.

adrianreber:
@Parthiba-Hazra Can you take a look add adding also an e2e test? Take a look at the `make test` and `make test-junit`.

Also, please to not include merge commits. Try to do a rebase of your PR using the latest `main` branch.

Parthiba-Hazra:
> @Parthiba-Hazra Can you take a look add adding also an e2e test? Take a look at the `make test` and `make test-junit`.
> 
> Also, please to not include merge commits. Try to do a rebase of your PR using the latest `main` branch.

@adrianreber yes I actually working on the test case and mistakenly update the branch .. sorry for that .

Parthiba-Hazra:
Hey, @adrianreber @rst0git I wrote some e2e tests, right now I am not pushing those changes as there will be some conflicts against the pr #116 .

list.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+
+				table.Append(row)
+			}(file)
+		}
+	}
+
+	wg.Wait()
+
+	table.Render()
```
rst0git, line 79: It would be better to render a table only when there are any checkpoints found.
For example, something like the following:
```diff
 @@ -18,6 +18,7 @@ import (
 
 func list(cmd *cobra.Command, args []string) error {
 	allPaths := append([]string{defaultCheckpointPath}, additionalCheckpointPaths...)
+	showTable := false
 
 	var wg sync.WaitGroup
 
@@ -35,7 +36,6 @@ func list(cmd *cobra.Command, args []string) error {
 	table.SetRowLine(true)
 
 	for _, checkpointPath := range allPaths {
-		fmt.Printf("Listing checkpoints in path: %s\n", checkpointPath)
 
 		files, err := filepath.Glob(filepath.Join(checkpointPath, "checkpoint-*"))
 		if err != nil {
@@ -43,10 +43,11 @@ func list(cmd *cobra.Command, args []string) error {
 		}
 
 		if len(files) == 0 {
-			fmt.Println("No checkpoints found in this path.")
 			continue
 		}
 
+		showTable = true
+		fmt.Printf("Listing checkpoints in path: %s\n", checkpointPath)
 		for _, file := range files {
 			wg.Add(1)
 
@@ -76,7 +77,11 @@ func list(cmd *cobra.Command, args []string) error {
 
 	wg.Wait()
 
-	table.Render()
+	if showTable {
+		table.Render()
+	} else {
+		fmt.Println("No checkpoints found")
+	}
 
 	return nil
 }
```

list.go
```diff
@@ -0,0 +1,100 @@
...
+		if len(files) == 0 {
+			fmt.Println("No checkpoints found in this path.")
+			continue
+		}
+
+		for _, file := range files {
+			wg.Add(1)
+
+			go func(file string) {
+				defer wg.Done()
```
rst0git, line 54: Could we simplify this to something like the following? This also formats the timestamp to make it easier to read.
```Go
for _, file := range files {
	namespace, pod, container, timestamp, err := parseCheckpointName(filepath.Base(file))
	if err != nil {
		log.Printf("Error parsing checkpoint name %s: %v\n", file, err)
		continue
	}

	timeValue, err := time.Parse(time.RFC3339, timestamp)
	if err != nil {
		log.Printf("Error parsing timestamp %s for file %s: %v\n", timestamp, file, err)
		continue
	}

	row := []string{
		namespace,
		pod,
		container,
		timeValue.Format(time.RFC822),
		filepath.Base(file),
	}

	table.Append(row)
}
```

test/checkpointctl.bats
```diff
@@ -589,3 +589,22 @@ function teardown() {
...
+    [ "$status" -eq 0 ]
+    [[ ${lines[1]} == *"No checkpoints found in this path."* ]]
+}
+
+@test "Run checkpointctl list with checkpoints in tar files" {
+    mkdir -p "$TEST_TMP_DIR1"/checkpoint-1
+    mkdir -p "$TEST_TMP_DIR1"/checkpoint-2
+
+    tar cf "$TEST_TMP_DIR2"/checkpoint-pod-name_namespace1-container-name-2023-10-12T11:02:56+01:00.tar -C "$TEST_TMP_DIR1" checkpoint-1
+    tar cf "$TEST_TMP_DIR2"/checkpoint-counter-tl58j_default-counter-2023-10-12T11:02:56+01:00.tar -C "$TEST_TMP_DIR1" checkpoint-2
```
rst0git, line 604: For this test we can use something like `touch <file path>` instead of `tar`.

test/checkpointctl.bats
```diff
@@ -589,3 +589,22 @@ function teardown() {
...
+
+    checkpointctl list -p "$TEST_TMP_DIR2"
+    [ "$status" -eq 0 ]
+    [[ "${lines[8]}" =~ "checkpoint-counter-tl58j_default-counter-2023-10-12T11:02:56+01:00.tar" ]]
+    [[ "${lines[6]}" =~ "checkpoint-pod-name_namespace1-container-name-2023-10-12T11:02:56+01:00.tar" ]]
```
rst0git, line 609: It would be good to verify that namespace, pod, and container are correctly parsed in the output table.

Parthiba-Hazra: currently I added assertion for other fields as same as the chekpoint name assertion in line 4 and 6. I think it could be done in a better way if we can parse those field's value in an array and then asserts them one by one using index value. is that possible? I am really not aware of that. anyway I will check it and if that possible then we can do in the better way.

test/checkpointctl.bats
```diff
@@ -589,3 +589,39 @@ function teardown() {
...
+
+    [[ "${lines[4]}" =~ "default" ]]
+    [[ "${lines[4]}" =~ "counter-tl58j" ]]
+    [[ "${lines[4]}" =~ "counter" ]]
+    [[ "${lines[4]}" =~ "12 Oct 23 11:02 +0100" ]]
+
+    [[ "${lines[6]}" =~ "namespace1" ]]
+    [[ "${lines[6]}" =~ "pod-name" ]]
+    [[ "${lines[6]}" =~ "container-name" ]]
+    [[ "${lines[6]}" =~ "12 Oct 23 11:02 +0100" ]]
```
snprajwal, line 620: Since they are complete substrings being matched, it would be better to use `==` with a wildcard rather than a regex match, e.g. `[[ "${lines[4]}" == *"default"* ]]`

list.go
```diff
@@ -0,0 +1,105 @@
...
+
+		if len(files) == 0 {
+			continue
+		}
+
+		showTable = true
+		fmt.Printf("Listing checkpoints in path: %s\n", checkpointPath)
+
+		for _, file := range files {
+			namespace, pod, container, timestamp, err := parseCheckpointName(filepath.Base(file))
```
adrianreber, line 54: I do not think parsing the file name is a good idea. It was not created to be parse-able and it might change in the future. It probably is faster to parse the name, but we have all the necessary information in the tar archive. Extracting a single file from the archive should give you much easier access to the necessary data.

rst0git: I agree, using data in the tar archive would be more reliable than parsing the file name. We can extract the fields `name` and `checkpointedTime` from `config.dump` (for checkpoints created with CRI-O). The field "name" contains the Kubernetes namespace, pod, and container names.

Parthiba-Hazra: @rst0git @adrianreber as far as I can see that from the config.dump file we can get the container name, image name, image id, ports, network mode, runtime, mounts etc. Although the checkpoint tar file that I am navigating is a podman container's, maybe I need to create a checkpoint of kubernetes pod.

rst0git: @Parthiba-Hazra Podman and CRI-O store different data in `config.dump`. Here is an example `config.dump` to illustrate this difference:
<details>
<summary>Podman</summary>

```json
{
  "spec": {
    "ociVersion": "1.1.0+dev",
    "process": {
      "user": {
        "uid": 0,
        "gid": 0
      },
      "args": [
        "/bin/sh",
        "-c",
        "i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done"
      ],
      "env": [
        "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin",
        "container=podman"
      ],
      "cwd": "/",
      "capabilities": {
        "bounding": [
          "CAP_CHOWN",
          "CAP_DAC_OVERRIDE",
          "CAP_FOWNER",
          "CAP_FSETID",
          "CAP_KILL",
          "CAP_NET_BIND_SERVICE",
          "CAP_SETFCAP",
          "CAP_SETGID",
          "CAP_SETPCAP",
          "CAP_SETUID",
          "CAP_SYS_CHROOT"
        ],
        "effective": [
          "CAP_CHOWN",
          "CAP_DAC_OVERRIDE",
          "CAP_FOWNER",
          "CAP_FSETID",
          "CAP_KILL",
          "CAP_NET_BIND_SERVICE",
          "CAP_SETFCAP",
          "CAP_SETGID",
          "CAP_SETPCAP",
          "CAP_SETUID",
          "CAP_SYS_CHROOT"
        ],
        "permitted": [
          "CAP_CHOWN",
          "CAP_DAC_OVERRIDE",
          "CAP_FOWNER",
          "CAP_FSETID",
          "CAP_KILL",
          "CAP_NET_BIND_SERVICE",
          "CAP_SETFCAP",
          "CAP_SETGID",
          "CAP_SETPCAP",
          "CAP_SETUID",
          "CAP_SYS_CHROOT"
        ]
      },
      "rlimits": [
        {
          "type": "RLIMIT_NPROC",
          "hard": 4194304,
          "soft": 4194304
        }
      ]
    },
    "root": {
      "path": "rootfs"
    },
    "mounts": [
      {
        "destination": "/proc",
        "type": "proc",
        "source": "proc",
        "options": [
          "nosuid",
          "noexec",
          "nodev"
        ]
      },
      {
        "destination": "/dev",
        "type": "tmpfs",
        "source": "tmpfs",
        "options": [
          "nosuid",
          "strictatime",
          "mode=755",
          "size=65536k"
        ]
      },
      {
        "destination": "/dev/pts",
        "type": "devpts",
        "source": "devpts",
        "options": [
          "nosuid",
          "noexec",
          "newinstance",
          "ptmxmode=0666",
          "mode=0620",
          "gid=5"
        ]
      },
      {
        "destination": "/dev/mqueue",
        "type": "mqueue",
        "source": "mqueue",
        "options": [
          "nosuid",
          "noexec",
          "nodev"
        ]
      },
      {
        "destination": "/sys",
        "type": "sysfs",
        "source": "sysfs",
        "options": [
          "nosuid",
          "noexec",
          "nodev",
          "ro"
        ]
      },
      {
        "destination": "/sys/fs/cgroup",
        "type": "cgroup",
        "source": "cgroup",
        "options": [
          "rprivate",
          "nosuid",
          "noexec",
          "nodev",
          "relatime",
          "ro"
        ]
      }
    ],
    "linux": {
      "sysctl": {
        "net.ipv4.ping_group_range": "0 0"
      },
      "resources": {
        "devices": [
          {
            "allow": false,
            "access": "rwm"
          }
        ],
        "pids": {
          "limit": 2048
        }
      },
      "namespaces": [
        {
          "type": "pid"
        },
        {
          "type": "network"
        },
        {
          "type": "ipc"
        },
        {
          "type": "uts"
        },
        {
          "type": "mount"
        },
        {
          "type": "cgroup"
        }
      ],
      "seccomp": {
        "defaultAction": "SCMP_ACT_ERRNO",
        "defaultErrnoRet": 38,
        "architectures": [
          "SCMP_ARCH_X86_64",
          "SCMP_ARCH_X86",
          "SCMP_ARCH_X32"
        ],
        "syscalls": [
          {
            "names": [
              "bdflush",
              "io_pgetevents",
              "kexec_file_load",
              "kexec_load",
              "migrate_pages",
              "move_pages",
              "nfsservctl",
              "nice",
              "oldfstat",
              "oldlstat",
              "oldolduname",
              "oldstat",
              "olduname",
              "pciconfig_iobase",
              "pciconfig_read",
              "pciconfig_write",
              "sgetmask",
              "ssetmask",
              "swapcontext",
              "swapoff",
              "swapon",
              "sysfs",
              "uselib",
              "userfaultfd",
              "ustat",
              "vm86",
              "vm86old",
              "vmsplice"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "_llseek",
              "_newselect",
              "accept",
              "accept4",
              "access",
              "adjtimex",
              "alarm",
              "bind",
              "brk",
              "capget",
              "capset",
              "chdir",
              "chmod",
              "chown",
              "chown32",
              "clock_adjtime",
              "clock_adjtime64",
              "clock_getres",
              "clock_getres_time64",
              "clock_gettime",
              "clock_gettime64",
              "clock_nanosleep",
              "clock_nanosleep_time64",
              "clone",
              "clone3",
              "close",
              "close_range",
              "connect",
              "copy_file_range",
              "creat",
              "dup",
              "dup2",
              "dup3",
              "epoll_create",
              "epoll_create1",
              "epoll_ctl",
              "epoll_ctl_old",
              "epoll_pwait",
              "epoll_pwait2",
              "epoll_wait",
              "epoll_wait_old",
              "eventfd",
              "eventfd2",
              "execve",
              "execveat",
              "exit",
              "exit_group",
              "faccessat",
              "faccessat2",
              "fadvise64",
              "fadvise64_64",
              "fallocate",
              "fanotify_mark",
              "fchdir",
              "fchmod",
              "fchmodat",
              "fchown",
              "fchown32",
              "fchownat",
              "fcntl",
              "fcntl64",
              "fdatasync",
              "fgetxattr",
              "flistxattr",
              "flock",
              "fork",
              "fremovexattr",
              "fsconfig",
              "fsetxattr",
              "fsmount",
              "fsopen",
              "fspick",
              "fstat",
              "fstat64",
              "fstatat64",
              "fstatfs",
              "fstatfs64",
              "fsync",
              "ftruncate",
              "ftruncate64",
              "futex",
              "futex_time64",
              "futimesat",
              "get_mempolicy",
              "get_robust_list",
              "get_thread_area",
              "getcpu",
              "getcwd",
              "getdents",
              "getdents64",
              "getegid",
              "getegid32",
              "geteuid",
              "geteuid32",
              "getgid",
              "getgid32",
              "getgroups",
              "getgroups32",
              "getitimer",
              "getpeername",
              "getpgid",
              "getpgrp",
              "getpid",
              "getppid",
              "getpriority",
              "getrandom",
              "getresgid",
              "getresgid32",
              "getresuid",
              "getresuid32",
              "getrlimit",
              "getrusage",
              "getsid",
              "getsockname",
              "getsockopt",
              "gettid",
              "gettimeofday",
              "getuid",
              "getuid32",
              "getxattr",
              "inotify_add_watch",
              "inotify_init",
              "inotify_init1",
              "inotify_rm_watch",
              "io_cancel",
              "io_destroy",
              "io_getevents",
              "io_setup",
              "io_submit",
              "ioctl",
              "ioprio_get",
              "ioprio_set",
              "ipc",
              "keyctl",
              "kill",
              "landlock_add_rule",
              "landlock_create_ruleset",
              "landlock_restrict_self",
              "lchown",
              "lchown32",
              "lgetxattr",
              "link",
              "linkat",
              "listen",
              "listxattr",
              "llistxattr",
              "lremovexattr",
              "lseek",
              "lsetxattr",
              "lstat",
              "lstat64",
              "madvise",
              "mbind",
              "membarrier",
              "memfd_create",
              "memfd_secret",
              "mincore",
              "mkdir",
              "mkdirat",
              "mknod",
              "mknodat",
              "mlock",
              "mlock2",
              "mlockall",
              "mmap",
              "mmap2",
              "mount",
              "mount_setattr",
              "move_mount",
              "mprotect",
              "mq_getsetattr",
              "mq_notify",
              "mq_open",
              "mq_timedreceive",
              "mq_timedreceive_time64",
              "mq_timedsend",
              "mq_timedsend_time64",
              "mq_unlink",
              "mremap",
              "msgctl",
              "msgget",
              "msgrcv",
              "msgsnd",
              "msync",
              "munlock",
              "munlockall",
              "munmap",
              "name_to_handle_at",
              "nanosleep",
              "newfstatat",
              "open",
              "open_tree",
              "openat",
              "openat2",
              "pause",
              "pidfd_getfd",
              "pidfd_open",
              "pidfd_send_signal",
              "pipe",
              "pipe2",
              "pivot_root",
              "pkey_alloc",
              "pkey_free",
              "pkey_mprotect",
              "poll",
              "ppoll",
              "ppoll_time64",
              "prctl",
              "pread64",
              "preadv",
              "preadv2",
              "prlimit64",
              "process_mrelease",
              "process_vm_readv",
              "process_vm_writev",
              "pselect6",
              "pselect6_time64",
              "ptrace",
              "pwrite64",
              "pwritev",
              "pwritev2",
              "read",
              "readahead",
              "readdir",
              "readlink",
              "readlinkat",
              "readv",
              "reboot",
              "recv",
              "recvfrom",
              "recvmmsg",
              "recvmmsg_time64",
              "recvmsg",
              "remap_file_pages",
              "removexattr",
              "rename",
              "renameat",
              "renameat2",
              "restart_syscall",
              "rmdir",
              "rseq",
              "rt_sigaction",
              "rt_sigpending",
              "rt_sigprocmask",
              "rt_sigqueueinfo",
              "rt_sigreturn",
              "rt_sigsuspend",
              "rt_sigtimedwait",
              "rt_sigtimedwait_time64",
              "rt_tgsigqueueinfo",
              "sched_get_priority_max",
              "sched_get_priority_min",
              "sched_getaffinity",
              "sched_getattr",
              "sched_getparam",
              "sched_getscheduler",
              "sched_rr_get_interval",
              "sched_rr_get_interval_time64",
              "sched_setaffinity",
              "sched_setattr",
              "sched_setparam",
              "sched_setscheduler",
              "sched_yield",
              "seccomp",
              "select",
              "semctl",
              "semget",
              "semop",
              "semtimedop",
              "semtimedop_time64",
              "send",
              "sendfile",
              "sendfile64",
              "sendmmsg",
              "sendmsg",
              "sendto",
              "set_mempolicy",
              "set_robust_list",
              "set_thread_area",
              "set_tid_address",
              "setfsgid",
              "setfsgid32",
              "setfsuid",
              "setfsuid32",
              "setgid",
              "setgid32",
              "setgroups",
              "setgroups32",
              "setitimer",
              "setns",
              "setpgid",
              "setpriority",
              "setregid",
              "setregid32",
              "setresgid",
              "setresgid32",
              "setresuid",
              "setresuid32",
              "setreuid",
              "setreuid32",
              "setrlimit",
              "setsid",
              "setsockopt",
              "setuid",
              "setuid32",
              "setxattr",
              "shmat",
              "shmctl",
              "shmdt",
              "shmget",
              "shutdown",
              "sigaction",
              "sigaltstack",
              "signal",
              "signalfd",
              "signalfd4",
              "sigpending",
              "sigprocmask",
              "sigreturn",
              "sigsuspend",
              "socket",
              "socketcall",
              "socketpair",
              "splice",
              "stat",
              "stat64",
              "statfs",
              "statfs64",
              "statx",
              "symlink",
              "symlinkat",
              "sync",
              "sync_file_range",
              "syncfs",
              "syscall",
              "sysinfo",
              "syslog",
              "tee",
              "tgkill",
              "time",
              "timer_create",
              "timer_delete",
              "timer_getoverrun",
              "timer_gettime",
              "timer_gettime64",
              "timer_settime",
              "timer_settime64",
              "timerfd",
              "timerfd_create",
              "timerfd_gettime",
              "timerfd_gettime64",
              "timerfd_settime",
              "timerfd_settime64",
              "times",
              "tkill",
              "truncate",
              "truncate64",
              "ugetrlimit",
              "umask",
              "umount",
              "umount2",
              "uname",
              "unlink",
              "unlinkat",
              "unshare",
              "utime",
              "utimensat",
              "utimensat_time64",
              "utimes",
              "vfork",
              "wait4",
              "waitid",
              "waitpid",
              "write",
              "writev"
            ],
            "action": "SCMP_ACT_ALLOW"
          },
          {
            "names": [
              "personality"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 0,
                "value": 0,
                "op": "SCMP_CMP_EQ"
              }
            ]
          },
          {
            "names": [
              "personality"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 0,
                "value": 8,
                "op": "SCMP_CMP_EQ"
              }
            ]
          },
          {
            "names": [
              "personality"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 0,
                "value": 131072,
                "op": "SCMP_CMP_EQ"
              }
            ]
          },
          {
            "names": [
              "personality"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 0,
                "value": 131080,
                "op": "SCMP_CMP_EQ"
              }
            ]
          },
          {
            "names": [
              "personality"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 0,
                "value": 4294967295,
                "op": "SCMP_CMP_EQ"
              }
            ]
          },
          {
            "names": [
              "arch_prctl"
            ],
            "action": "SCMP_ACT_ALLOW"
          },
          {
            "names": [
              "modify_ldt"
            ],
            "action": "SCMP_ACT_ALLOW"
          },
          {
            "names": [
              "open_by_handle_at"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "bpf",
              "fanotify_init",
              "lookup_dcookie",
              "perf_event_open",
              "quotactl",
              "setdomainname",
              "sethostname",
              "setns"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "chroot"
            ],
            "action": "SCMP_ACT_ALLOW"
          },
          {
            "names": [
              "delete_module",
              "finit_module",
              "init_module",
              "query_module"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "acct"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "kcmp",
              "process_madvise"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "ioperm",
              "iopl"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "clock_settime",
              "clock_settime64",
              "settimeofday",
              "stime"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "vhangup"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 1
          },
          {
            "names": [
              "socket"
            ],
            "action": "SCMP_ACT_ERRNO",
            "errnoRet": 22,
            "args": [
              {
                "index": 0,
                "value": 16,
                "op": "SCMP_CMP_EQ"
              },
              {
                "index": 2,
                "value": 9,
                "op": "SCMP_CMP_EQ"
              }
            ]
          },
          {
            "names": [
              "socket"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 2,
                "value": 9,
                "op": "SCMP_CMP_NE"
              }
            ]
          },
          {
            "names": [
              "socket"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 0,
                "value": 16,
                "op": "SCMP_CMP_NE"
              }
            ]
          },
          {
            "names": [
              "socket"
            ],
            "action": "SCMP_ACT_ALLOW",
            "args": [
              {
                "index": 2,
                "value": 9,
                "op": "SCMP_CMP_NE"
              }
            ]
          }
        ]
      },
      "maskedPaths": [
        "/proc/acpi",
        "/proc/kcore",
        "/proc/keys",
        "/proc/latency_stats",
        "/proc/timer_list",
        "/proc/timer_stats",
        "/proc/sched_debug",
        "/proc/scsi",
        "/sys/firmware",
        "/sys/fs/selinux",
        "/sys/dev/block"
      ],
      "readonlyPaths": [
        "/proc/asound",
        "/proc/bus",
        "/proc/fs",
        "/proc/irq",
        "/proc/sys",
        "/proc/sysrq-trigger"
      ]
    }
  },
  "id": "1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06",
  "name": "looper",
  "lockID": 16,
  "CreateCommand": [
    "podman",
    "run",
    "-d",
    "--name",
    "looper",
    "busybox",
    "/bin/sh",
    "-c",
    "i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done"
  ],
  "RawImageName": "busybox",
  "idMappingsOptions": {
    "HostUIDMapping": true,
    "HostGIDMapping": true,
    "UIDMap": null,
    "GIDMap": null,
    "AutoUserNs": false,
    "AutoUserNsOpts": {
      "Size": 0,
      "InitialSize": 0,
      "PasswdFile": "",
      "GroupFile": "",
      "AdditionalUIDMappings": null,
      "AdditionalGIDMappings": null
    }
  },
  "Dependencies": null,
  "rootfsImageID": "9211bbaa0dbd68fed073065eb9f0a6ed00a75090a9235eca2554c62d1e75c58f",
  "rootfsImageName": "docker.io/library/busybox:latest",
  "ShmDir": "/var/lib/containers/storage/overlay-containers/1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06/userdata/shm",
  "shmSize": 65536000,
  "shmSizeSystemd": 0,
  "staticDir": "/var/lib/containers/storage/overlay-containers/1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06/userdata",
  "mounts": [
    "/var/lib/containers/storage/overlay-containers/1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06/userdata/shm"
  ],
  "createWorkingDir": true,
  "secretsPath": "/var/lib/containers/storage/overlay-containers/1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06/userdata/secrets",
  "storageOpts": null,
  "passwd": true,
  "privileged": false,
  "ProcessLabel": "system_u:system_r:container_t:s0:c569,c819",
  "MountLabel": "system_u:object_r:container_file_t:s0:c569,c819",
  "label_nested": false,
  "createNetNS": true,
  "UseImageResolvConf": false,
  "UseImageHosts": false,
  "newNetworks": {
    "podman": {
      "interface_name": "eth0"
    }
  },
  "networkMode": "bridge",
  "command": [
    "/bin/sh",
    "-c",
    "i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done"
  ],
  "stopSignal": 15,
  "stopTimeout": 10,
  "createdTime": "2024-01-19T09:35:21.429939638Z",
  "cgroupManager": "systemd",
  "cgroupsMode": "enabled",
  "cgroupParent": "machine.slice",
  "kubeExitCodePropagation": 0,
  "logPath": "",
  "logTag": "",
  "logSize": 0,
  "logDriver": "journald",
  "conmonPidFile": "/run/containers/storage/overlay-containers/1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06/userdata/conmon.pid",
  "postConfigureNetNS": false,
  "runtime": "crun",
  "pause": false,
  "isService": false,
  "sdnotifyMode": "container",
  "healthcheck": null,
  "healthcheck_on_failure_action": 0,
  "umask": "0022",
  "pid_file": "/run/containers/storage/overlay-containers/1a92a6937e4be4fdd64832792bbd5fb4a92f96a3f95be19a8d48a6272068ec06/userdata/pidfile",
  "mountAllDevices": false,
  "readWriteTmpfs": false
}
```
</details>

<details>
<summary>CRI-O</summary>

```json
{
  "id": "f3d1eb1107d2b78de5e445303b753569a814d9973443cc8e792580997f11c352",
  "name": "k8s_counter_counter-gvb5j_default_ee0dc0b9-d4de-4641-9f3f-554b72d4390e_0",
  "rootfsImage": "8c811b4aec35f259572d0f79207bc0678df4c736eeec50bc9fec37ed936a472a",
  "rootfsImageRef": "8c811b4aec35f259572d0f79207bc0678df4c736eeec50bc9fec37ed936a472a",
  "rootfsImageName": "docker.io/library/busybox:1.28",
  "runtime": "runc",
  "createdTime": "2023-10-11T09:36:24.968326678Z",
  "checkpointedTime": "2023-10-12T11:01:57.051768645+01:00",
  "restoredTime": "0001-01-01T00:00:00Z",
  "restored": false
}
```
</details>

Podman (similar to Docker) is mainly used to create and manage containers and it might be more useful for the `checkpointctl list` command to show an overview of container checkpoints that have not been exported, i.e., stored in `/var/lib/containers/storage/overlay-containers/`. Currently, there is no easy way of finding out what checkpoints have been created other than 

```
ls /var/lib/containers/storage/overlay-containers/*/userdata/checkpoint
```


On the other hand, CRI-O is similar to [containerd](https://github.com/containerd/containerd). With CRI-O, the `checkpointctl list` command is somewhat similar to `checkpointctl show /var/lib/kubelet/checkpoints/*`. The difference is that it shows information about the Kubernetes namespace and Pod for each checkpoint, and we can extend the list command with options to filter the results. This is particularly useful when container checkpoints are created automatically.

Parthiba-Hazra: @rst0git thanks for the clarification, I also thought the same. Actually I was having some trouble to install CRI-O, that's why I was not tested that.

Parthiba-Hazra: @adrianreber @rst0git I need one help from you. 
I started minikube using crio runtime -
```
➜ minikube start --container-runtime=crio --feature-gates=ContainerCheckpoint=true
```
applied a demo pod - 
```
➜ kubectl apply -f demo-pod.yaml
```
demo-pod.yaml- 
```
apiVersion: v1
kind: Pod
metadata:
  name: busybox-pod
spec:
  containers:
  - name: busybox-container
    image: busybox
    command: ["sleep", "3600"]
```
but creating a checkpoint was not successful - 
```
docker@minikube:~$ sudo crictl checkpoint busybox-container
E0120 15:37:07.591731   12262 remote_runtime.go:1214] "CheckpointContainer from runtime service failed" err="rpc error: code = Unimplemented desc = unknown method CheckpointContainer for service runtime.v1.RuntimeService" containerID="busybox-container"
FATA[0000] checkpointing the container "busybox-container" failed: rpc error: code = Unimplemented desc = unknown method CheckpointContainer for service runtime.v1.RuntimeService 

```

Parthiba-Hazra: also tried - 
```
➜  list-test curl --insecure --cert ~/.minikube/profiles/criupoc/client.crt --key ~/.minikube/profiles/criupoc/client.key -X POST "https://<minikube ip>:10250/checkpoint/default/busybox-pod/busybox-container"
curl: (28) Failed to connect to <minikube ip> port 10250 after 134275 ms: Couldn't connect to server

```

Parthiba-Hazra: this also fail- 
```
docker@minikube:~$ vi client.crt
docker@minikube:~$ vi client.key
docker@minikube:~$ curl -skv --cert ./client.crt --key ./client.key -X POST "https://localhost:10250/checkpoint/default/busybox-pod/busybox-container"
*   Trying 127.0.0.1:10250...
* Connected to localhost (127.0.0.1) port 10250 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* TLSv1.0 (OUT), TLS header, Certificate Status (22):
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.2 (IN), TLS header, Certificate Status (22):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.2 (IN), TLS header, Finished (20):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Request CERT (13):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.2 (OUT), TLS header, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.3 (OUT), TLS handshake, Certificate (11):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.3 (OUT), TLS handshake, CERT verify (15):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_128_GCM_SHA256
* ALPN, server accepted to use h2
* Server certificate:
*  subject: CN=minikube@1705764396
*  start date: Jan 20 14:26:36 2024 GMT
*  expire date: Jan 19 14:26:36 2025 GMT
*  issuer: CN=minikube-ca@1705764396
*  SSL certificate verify result: self-signed certificate in certificate chain (19), continuing anyway.
* Using HTTP2, server supports multiplexing
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* Using Stream ID: 1 (easy handle 0x55ad97ca0e90)
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
> POST /checkpoint/default/busybox-pod/busybox-container HTTP/2
> Host: localhost:10250
> user-agent: curl/7.81.0
> accept: */*
> 
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
* TLSv1.2 (OUT), TLS header, Supplemental data (23):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
* TLSv1.2 (IN), TLS header, Supplemental data (23):
< HTTP/2 500 
< content-type: text/plain; charset=utf-8
< content-length: 175
< date: Sat, 20 Jan 2024 17:02:15 GMT
< 
* Connection #0 to host localhost left intact
checkpointing of default/busybox-pod/busybox-container failed (rpc error: code = Unimplemented desc = unknown method CheckpointContainer for service runtime.v1.Runt
```

Parthiba-Hazra: @rst0git need some more time to update this PR, still having some issue to run kubeadm in my ubuntu machine ;(

internal/config_extractor.go
```diff
@@ -0,0 +1,71 @@
...
+	content, err := os.ReadFile(configDumpPath)
+	if err != nil {
+		log.Printf("Error reading config.dump file: %v\n", err)
+		return "", "", "", time.Time{}, err
+	}
+
+	return extractConfigDumpContent(content)
+}
+
+type Content struct {
```
adrianreber, line 36: You should use https://github.com/checkpoint-restore/checkpointctl/blob/main/lib/metadata.go#L38

Parthiba-Hazra:
@adrianreber @rst0git need a review on this pr :)

adrianreber:
The check binary size CI run fails because one of the intermediate commits does not build. Is this one of your commits or was one of the previous commits in the repository broken?

adrianreber:
Please also take a look at your commit messages. They do not contain much information why you made the changes you did.

docs/checkpointctl-list.adoc
```diff
@@ -0,0 +1,22 @@
+= checkpointctl-list(1)
+include::footer.adoc[]
+
+== Name
+
+*checkpointctl-list* - List checkpoints stored in the default and additional paths
+
```
adrianreber, line 7: Please list the default location of the path in the documentation.

Parthiba-Hazra:
> The check binary size CI run fails because one of the intermediate commits does not build. Is this one of your commits or was one of the previous commits in the repository broken?

actually that is one of previous commit . should I squash them?

internal/config_extractor.go
Parthiba-Hazra: yes I update that.

internal/config_extractor.go
```diff
@@ -0,0 +1,55 @@
...
+	return extractConfigDumpContent(content)
+}
+
+func extractConfigDumpContent(content []byte) (namespace, pod, container string, timestamp time.Time, err error) {
+	var c metadata.ContainerConfig
+	if err := json.Unmarshal(content, &c); err != nil {
+		return "", "", "", time.Time{}, err
+	}
+
+	parts := strings.Split(c.Name, "_")
```
adrianreber, line 44: I would prefer to avoid string parsing. This information is available from the file spec.dump in the archive. You could then also add a field which container engine created the checkpoint. Currently it is only CRI-O for Kubernetes, but there is a PR for containerd open (for a long time).

In the annotations of spec.dump there is the information in io.kubernetes.cri-o.Labels. Take a look at https://github.com/cri-o/cri-o/blob/main/server/container_restore.go#L162

Parthiba-Hazra: Got it. Thanks @adrianreber for the information, I should also check the spec.dump. Sorry for the inconvenience.

Parthiba-Hazra:
@rst0git @adrianreber I updated the PR with desired changes :)

adrianreber:
From my point of view the commits need to be cleaned up. The commit messages seem to be the result of multiple squashes. Only one Signed-off-by is necessary and useful commit message please. Try to break lines after 72 characters.

There are multiple guides about how to write good messages. I think @rst0git always has a good link but I recently read this one (https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines) which also did not sound too bad.

In one commit you remove an unrelated test and in another commit you remove it again. That looks unnecessary. The changes also seem to be distributed in a strange way across multiple commits. It makes sense to split different changes into different commits, but not like in this PR. I think the structure of the commits needs more work at this point.

Parthiba-Hazra:
> From my point of view the commits need to be cleaned up. The commit messages seem to be the result of multiple squashes. Only one Signed-off-by is necessary and useful commit message please. Try to break lines after 72 characters.
> 
> There are multiple guides about how to write good messages. I think @rst0git always has a good link but I recently read this one (https://github.com/kubernetes/community/blob/master/contributors/guide/pull-requests.md#commit-message-guidelines) which also did not sound too bad.
> 
> In one commit you remove an unrelated test and in another commit you remove it again. That looks unnecessary. The changes also seem to be distributed in a strange way across multiple commits. It makes sense to split different changes into different commits, but not like in this PR. I think the structure of the commits needs more work at this point.

You are right currently the commits are really messed up. let me work on this.. if necessary I can create a new pr with proper commits..

adrianreber:
> You are right currently the commits are really messed up. let me work on this.. if necessary I can create a new pr with proper commits.

This is usually never necessary. With the help of `git rebase` and `git push --force` to your branch it should not be necessary to have a new PR. Also use `git pull --rebase` to rebase your work on the latest `main` branch.

Parthiba-Hazra:
> > You are right currently the commits are really messed up. let me work on this.. if necessary I can create a new pr with proper commits.
> 
> This is usually never necessary. With the help of `git rebase` and `git push --force` to your branch it should not be necessary to have a new PR. Also use `git pull --rebase` to rebase your work on the latest `main` branch.

yea that's what I am gonna do. will try to fix this soon.

Parthiba-Hazra:
@adrianreber @rst0git I hope now it satisfy the **commit message guidelines** .

snprajwal:
In order for us to use `git bisect` while debugging regressions, we require all commits to be self-contained, i.e. every commit must be able to successfully build and pass tests. It looks like your first commit has changes that are actually a part of the second commit, leading to the build breaking. That is what is causing the CI to fail - the logs indicate what the Go build errors are, and in which file.

Parthiba-Hazra:
> In order for us to use `git bisect` while debugging regressions, we require all commits to be self-contained, i.e. every commit must be able to successfully build and pass tests. It looks like your first commit has changes that are actually a part of the second commit, leading to the build breaking. That is what is causing the CI to fail - the logs indicate what the Go build errors are, and in which file.

That's right, I really messed up things while solving merge conflicts. Really sorry for this .

rst0git:
> That's right, I really messed up things while solving merge conflicts. Really sorry for this .

No worries, don't take code reviews as criticism; they are just advice on how your pull request can be improved ;)

rst0git (approved):
LGTM

internal/utils.go
```diff
@@ -82,3 +85,54 @@ func CleanupTasks(tasks []Task) {
 		}
 	}
 }
+
+func CreateTarWithFiles(tarPath string, filePaths []string) error {
```
snprajwal, line 89: This function is being used exclusively for tests - I don't think it belongs in this file since it is not a general utility function

cmd/list.go
```diff
@@ -0,0 +1,101 @@
...
+	allPaths := append([]string{defaultCheckpointPath}, additionalCheckpointPaths...)
+	showTable := false
+
+	table := tablewriter.NewWriter(os.Stdout)
+	header := []string{
+		"Namespace",
+		"Pod",
+		"Container",
+		"Engine",
+		"Time Created",
```
snprajwal, line 53: We can also add a "Time Checkpointed" field here, thanks to #120

internal/config_extractor_test.go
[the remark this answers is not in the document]
Parthiba-Hazra: I think, I just need to remove `TestExtractConfigDump` function and `config.dump` and `spec.dump` files are already placed under test/data/list_config_spec.dump by this PR.

cmd/list.go
Parthiba-Hazra: Ofcourse we can, but the problem here is that we already have so many columns for the table and it already became too messy while even the terminal is open with full display, should we print it in a different format? 
like - 
```
Checkpoint Name: 
Time Checkpointed:
Time Created:
Namespace:
Pod:
Container:
Engine:
-----------------------------some ending pattern--------------------
```

snprajwal: Fair point, what do you think @adrianreber @rst0git

rst0git: > We can also add a "Time Checkpointed" field here

Actually, the current implementation shows the value of `config.CheckpointedAt`. This value indicates the time when the checkpoint was created, and it is also included in the file name.

Example:
```
$ checkpointctl list
Listing checkpoints in path: /var/lib/kubelet/checkpoints/
+-----------+---------------+-----------+--------+---------------------+------------------------------------------------------------------------+
| NAMESPACE |      POD      | CONTAINER | ENGINE |    TIME CREATED     |                            CHECKPOINT NAME                             |
+-----------+---------------+-----------+--------+---------------------+------------------------------------------------------------------------+
| default   | counter-ccpp2 | counter   | cri-o  | 12 Oct 23 11:01 BST | checkpoint-counter-ccpp2_default-counter-2023-10-12T11:01:56+01:00.tar |
+-----------+---------------+-----------+--------+---------------------+------------------------------------------------------------------------+
| default   | counter-ccpp2 | counter   | cri-o  | 12 Oct 23 11:02 BST | checkpoint-counter-ccpp2_default-counter-2023-10-12T11:02:54+01:00.tar |
+-----------+---------------+-----------+--------+---------------------+------------------------------------------------------------------------+
```

`config.CreatedTime` indicates the time when the container was created, i.e., if there are multiple checkpoints of a container, all checkpoints would have the same `CreatedTime` value.

snprajwal (approved):

rst0git:
@adrianreber PTAL

test/checkpointctl.bats
```diff
@@ -640,3 +640,55 @@ function teardown() {
...
+	touch "$TEST_TMP_DIR1"/spec.dump
+	cp data/list_config_spec.dump/config.dump "$TEST_TMP_DIR1"
+	mkdir "$TEST_TMP_DIR1"/checkpoint
+	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/checkpoint-config.tar . )
+	checkpointctl list -p "$TEST_TMP_DIR2"
+	[ "$status" -eq 0 ]
+	[[ ${lines[1]} == *"Error extracting information from $TEST_TMP_DIR2/checkpoint-config.tar: unexpected end of JSON input" ]]
+}
+
+@test "Run checkpointctl list with tar file with valid config.dump nad spec.dump" {
```
adrianreber, line 685: Can you create two checkpoint archives to run the list command against. Just to see that we see two checkpoints and not just a single one. You could use `sed` or `jq` to change one of the JSON fields to see two different results.

Parthiba-Hazra: yes sure.

adrianreber:
The commit structure still needs to be reworked from my point of view.

You have two commits. First commit is:

* `feat: Add support for the list command to retrieve checkpoints `: This looks correct from my point of view.
* `refactor: Extract necessary values from spec.dump`: This is the problematic one from my point of view:
  * All changes to the `list` command should happen to the first commit. My point of view is, if you introduce a feature in a single PR no fixup commits should be necessary. You restructure code in the second commit. You remove complete functions (`CreateTarWithFile`). All that should happen in the first commit. Do not introduce something and then change it in the next commit. As long as it is not part of the main repository you can change the commit structure as much as you want. Once it has been committed and we figure out it needs a change, a commit is the right thing to do. But not in a single PR, as long as it is all contained in a PR no need for a second refactor commit. Always, just from my point of view.
  * For me, in this case it means that everything could be in one commit. I would try to separate code changes and changes to `test/` in two commits, but this is something I would not insist on. It would be just nicer for a PR of this size to have separate commits for code and test changes. For a small commit it could all be in one commit.

If the `refactor` commit would touch other code, then that commit should be first and then your changes. But all the refactoring, as far as I can tell, is happening in the code from the previous commit, right?

As always, most things I mention are not absolute rules and it depends on the context.

Parthiba-Hazra:
> The commit structure still needs to be reworked from my point of view.
> 
> You have two commits. First commit is:
> 
> * `feat: Add support for the list command to retrieve checkpoints `: This looks correct from my point of view.
> * `refactor: Extract necessary values from spec.dump`: This is the problematic one from my point of view:
>   
>   * All changes to the `list` command should happen to the first commit. My point of view is, if you introduce a feature in a single PR no fixup commits should be necessary. You restructure code in the second commit. You remove complete functions (`CreateTarWithFile`). All that should happen in the first commit. Do not introduce something and then change it in the next commit. As long as it is not part of the main repository you can change the commit structure as much as you want. Once it has been committed and we figure out it needs a change, a commit is the right thing to do. But not in a single PR, as long as it is all contained in a PR no need for a second refactor commit. Always, just from my point of view.
>   * For me, in this case it means that everything could be in one commit. I would try to separate code changes and changes to `test/` in two commits, but this is something I would not insist on. It would be just nicer for a PR of this size to have separate commits for code and test changes. For a small commit it could all be in one commit.
> 
> If the `refactor` commit would touch other code, then that commit should be first and then your changes. But all the refactoring, as far as I can tell, is happening in the code from the previous commit, right?
> 
> As always, most things I mention are not absolute rules and it depends on the context.

@adrianreber I understand. I will squash the second commit and keep the PR in one commit. Thank you for the detailed explanation; I was not really aware of good commit practices before this PR. ;)

rst0git (approved):

PR #116, opened 2024-01-10
test: add e2e test for json format

rst0git:
This pull request adds a test case that verifies the output JSON format for checkpointctl.

### DIFF
--- a/test/checkpointctl.bats
+++ b/test/checkpointctl.bats
@@ -359,9 +359,9 @@ function teardown() {
 	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/test.tar . )
 	checkpointctl inspect "$TEST_TMP_DIR2"/test.tar --files
 	[ "$status" -eq 0 ]
-	[[ ${lines[11]} == *"[REG 0]"* ]]
-	[[ ${lines[12]} == *"[cwd]"* ]]
-	[[ ${lines[13]} == *"[root]"* ]]
+	[[ ${lines[24]} == *"[REG 0]"* ]]
+	[[ ${lines[25]} == *"[cwd]"* ]]
+	[[ ${lines[26]} == *"[root]"* ]]
 }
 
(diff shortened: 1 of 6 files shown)


rst0git:
cc: @Parthiba-Hazra

snprajwal (approved):

adrianreber:
I see a coverage of 20% of `json.go`. Would a higher coverage require a more complex test process?

rst0git:
> Would a higher coverage require a more complex test process?

Not necessarily. I will update the patch to extend the test coverage.

rst0git:
@adrianreber I've updated the pull request to increase test coverage.

test/checkpointctl.bats
```diff
@@ -14,7 +14,15 @@ function checkpointctl() {
...
 }
 
+run_only_test() {
+	if [ "$BATS_TEST_NUMBER" -ne "$1" ]; then
+		skip
+	fi
+}
+
 function setup() {
+	# run_only_test 49
```
adrianreber, line 24: Is this to only run certain tests? I was always using

`-f, --filter <regex>      Only run tests that match the regular expression`

To only run one test.

test/piggie/piggie.c
```diff
@@ -34,6 +144,9 @@ static int do_test(void *logf)
 	if (fd != 1 && fd != 2)
 		close(fd);
 
+	run_tcp_server();
```
adrianreber, line 147: Maybe we could only run the TCP part for selected tests. I am bit worried that always opening port 5000 will result in some problems sometimes. Port already in use or something. Alternatively you could try to open the port in a network namespace which would also include namespaces in the test binary.

test/checkpointctl.bats
rst0git: Thanks! I've removed this function.

test/piggie/piggie.c
rst0git: Good point! I've updated the test to create a network namepsace and use command-line options for TCP sockets and log file.

adrianreber (approved):

PR #117, opened 2024-01-11
Refactor: Split commands into separate package

### LINKED ISSUE #80: Refactor `checkpointctl.go` and move commands into individual files
The `checkpointctl.go` file has grown after adding the new `inspect` command, and will grow further as we expand the CLI. It would be better to refactor it into it's own package, and use the existing file only to invoke the main function.

shikharish:
Closes #80.


shikharish:
@snprajwal @adrianreber Any reviews?

internal/tree.go
```diff
@@ -12,7 +12,23 @@ import (
 	"github.com/xlab/treeprint"
 )
 
-func renderTreeView(tasks []task) error {
+var (
+	Name           string
+	Version        string
```
snprajwal, line 17: These variables are provided through the Makefile, and are not used in any of the commands - only in `checkpointctl version`. It would be better to have this in `checkpointctl.go`, since it has nothing to do with the tree view.

internal/tree.go
```diff
@@ -12,7 +12,23 @@ import (
...
+	Stats          bool
+	Mounts         bool
+	OutputFilePath string
+	PID            uint32
+	PsTree         bool
+	PsTreeCmd      bool
+	PsTreeEnv      bool
+	Files          bool
+	Sockets        bool
+	ShowAll        bool
```
snprajwal, line 28: These are common to both the tree and JSON views, so it would make more sense to have it in a separate file (possibly `options.go`?)

shikharish: Should I do the same for the cmd package where the variables are used by all three commands?

snprajwal:
@shikharish can you please squash all the commits into a single one? The first commit not being signed-off triggers the DCO failure. It would also be helpful to rename the PR to something like `refactor: split commands into separate package`

shikharish:
@snprajwal Any more reviews?

snprajwal:
@rst0git @adrianreber PTAL

snprajwal:
Also, @shikharish you would have to rebase this branch against the main branch of the repo, some commits have been added in the past week

rst0git:
@shikharish Would you be able to add a git commit message describing what changes have been made?
It would make it much easier to review ;-)

rst0git:
> Should I make a more detailed commit message?

@shikharish It is a good practice to do that. The following contributing guidelines and blog post provide more information: 
- https://github.com/checkpoint-restore/criu/blob/criu-dev/CONTRIBUTING.md#describe-your-changes
- https://cbea.ms/git-commit/

shikharish:
@rst0git Updated commit message. I hope this is better. Thank you for the informative resources.

rst0git (approved):
@shikharish Well done!

adrianreber:
Looks good indeed. All tests look happy.

PR #118, opened 2024-01-26
make: generate auto-completion files

rst0git:
This pull request extends the Makefile with targets to generate, install, and uninstall auto-completion files.


adrianreber:
Do we need to keep the generated files in the repository?

rst0git:
> Do we need to keep the generated files in the repository?

I used an approach similar to Podman (https://github.com/containers/podman/tree/main/completions); it simplifies the installation process.

adrianreber:
I kind of like it to have information about the completion in the main README file. No need from my side to move it somewhere else. Besides that, this looks good.

PR #120, opened 2024-02-09
Handle time stamps from containerd checkpoint archives

adrianreber:
This fixes containerd archive created time handling and adds checkpointed time for all archives.

### DIFF
--- a/internal/json.go
+++ b/internal/json.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/checkpoint-restore/go-criu/v7/crit"
 	spec "github.com/opencontainers/runtime-spec/specs-go"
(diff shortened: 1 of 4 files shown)


rst0git:
@adrianreber Would it be possible to also update the examples in [README.md](https://github.com/checkpoint-restore/checkpointctl/blob/main/README.md)?

adrianreber:
> @adrianreber Would it be possible to also update the examples in [README.md](https://github.com/checkpoint-restore/checkpointctl/blob/main/README.md)?

Sure. Done.

internal/json.go
```diff
@@ -112,6 +114,7 @@ func RenderJSONView(tasks []Task) error {
 			ID:            info.configDump.ID,
 			Runtime:       info.configDump.OCIRuntime,
 			Created:       info.containerInfo.Created,
+			Checkpointed:  info.configDump.CheckpointedAt.Format(time.RFC3339),
```
rst0git, line 117: It looks like Podman currently does not set `CheckpointedAt`

adrianreber: Interesting. We should set that. I will change this PR to skip it if empty.

adrianreber: Removed the zero value if not set as in the Podman case.

rst0git (approved):
LGTM

PR #122, opened 2024-03-06
Simplified 'list' command code

adrianreber:
The code for the 'list' command introduced code which already existed in the repository. This commit re-uses existing code.

Also, this extends the 'list' command to get the information from checkpoints created by containerd.

This also changes how the 'list' command works. Before it was always using '/var/lib/kubelet/checkpoints/' to look for checkpoints and '-p' extended the search path. Now 'checkpoint list' looks in the default location and 'checkpoint list /tmp /etc' only looks in '/tmp/' and '/etc', but no longer in the default location. Always looking in the default location broke local tests for me.

I am combining tests and code changes in a single commit as the changes are breaking the tests.

@Parthiba-Hazra The main reason I did the changes was that the information was not correctly displayed for the checkpoint archives as created by containerd.


rst0git (approved):
LGTM
Thanks Adrian!

PR #125, opened 2024-04-03
Add support to convert checkpoint archives to OCI images

### LINKED ISSUE #52: Extend checkpointctl to convert checkpoint archives to OCI images
Currently it is necessary to run a couple of `buildah` commands to convert a checkpoint archive to an OCI image which can be used by CRI-O or Podman.

It should be possible to do those steps in `checkpointctl` with one single command. If this is implemented it is important that including `lib/` in another go project does not pull in any additional dependencies (like buildah).

Parthiba-Hazra:
Resolves: #52 

This PR introduce a new command `checkpointctl build`, which can be used to convert checkpoint archives to OCI images. The command accepts a checkpoint archive and a image name as input and generates an OCI image suitable for use with container runtimes like CRI-O or Podman.

```
➜  ~ checkpointctl build /var/lib/kubelet/checkpoints/checkpoint-nginx-deployment-85bfd45d84-kfxmm_default-nginx-2024-03-29T11:32:17+05:30.tar test-checkpointctl:v1 

2024/04/03 19:05:26 Created checkpoint image: 7b3135033935815023dcdc810419ee1894621436b4d7622440691e4197db451c
2024/04/03 19:05:26 Image 'test-checkpointctl:v1' created successfully from checkpoint '/var/lib/kubelet/checkpoints/checkpoint-nginx-deployment-85bfd45d84-kfxmm_default-nginx-2024-03-29T11:32:17+05:30.tar'
```
```
➜  ~ podman images
REPOSITORY                                 TAG              IMAGE ID      CREATED         SIZE
docker.io/library/test-checkpointctl       v1               7b3135033935  33 seconds ago  15.8 MB

```

we can inspect the image using podman - 
```
➜  ~  podman image inspect 7b3135033935                           
[
     {
          "Id": "7b3135033935815023dcdc810419ee1894621436b4d7622440691e4197db451c",
          "Digest": "sha256:08ee0bc5036915f52b4a2d7a1a9746302cd4010322ef609b121407589b476078",
          "RepoTags": [
               "docker.io/library/test-checkpointctl:v1"
          ],
          "RepoDigests": [
               "docker.io/library/test-checkpointctl@sha256:08ee0bc5036915f52b4a2d7a1a9746302cd4010322ef609b121407589b476078"
          ],
          "Parent": "",
          "Comment": "",
          "Created": "2024-04-03T13:35:26.175477795Z",
          "Config": {},
          "Version": "",
          "Author": "",
          "Architecture": "amd64",
          "Os": "linux",
          "Size": 15841874,
          "VirtualSize": 15841874,
          "GraphDriver": {
               "Name": "overlay",
               "Data": {
                    "UpperDir": "/var/lib/containers/storage/overlay/fc8fdb8e1d92bfb6fd9d5e448a90993389fc3837f5ba3e226602d3c7b6de2e71/diff",
                    "WorkDir": "/var/lib/containers/storage/overlay/fc8fdb8e1d92bfb6fd9d5e448a90993389fc3837f5ba3e226602d3c7b6de2e71/work"
               }
          },
          "RootFS": {
               "Type": "layers",
               "Layers": [
                    "sha256:fc8fdb8e1d92bfb6fd9d5e448a90993389fc3837f5ba3e226602d3c7b6de2e71"
               ]
          },
          "Labels": null,
          "Annotations": {
               "io.container.manager": "CRI-O",
               "io.kubernetes.cri-o.annotations.checkpoint.name": "nginx",
               "io.kubernetes.cri-o.annotations.checkpoint.namespace": "default",
               "io.kubernetes.cri-o.annotations.checkpoint.pod": "nginx-deployment-85bfd45d84-kfxmm",
               "io.kubernetes.cri-o.annotations.checkpoint.rootfsImage": "docker.io/library/nginx@sha256:52478f8cd6a142fd462f0a7614a7bb064e969a4c083648235d6943c786df8cc7",
               "io.kubernetes.cri-o.annotations.checkpoint.rootfsImageID": "92b11f67642b62bbb98e7e49169c346b30e20cd3c1c034d31087e46924b9312e",
               "io.kubernetes.cri-o.annotations.checkpoint.rootfsImageName": "docker.io/library/nginx:latest",
               "io.kubernetes.cri-o.annotations.checkpoint.runtime.name": "runc",
               "org.opencontainers.image.base.digest": "",
               "org.opencontainers.image.base.name": ""
          },
          "ManifestType": "application/vnd.oci.image.manifest.v1+json",
          "User": "",
          "History": [
               {
                    "created": "2024-04-03T13:35:26.253908385Z",
                    "created_by": "/bin/sh"
               }
          ],
          "NamesHistory": [
               "docker.io/library/test-checkpointctl:v1"
          ]
     }
]

```

### DIFF
--- a/internal/scripts/build_image.sh
+++ b/internal/scripts/build_image.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+set -euo pipefail
+
+usage() {
+	cat <<EOF
+Usage: ${0##*/} [-a ANNOTATIONS_FILE] [-c CHECKPOINT_PATH] [-i IMAGE_NAME]
+Create OCI image from a checkpoint tar file.
+
+    -a          path to the annotations file
+    -c          path to the checkpoint file
+    -i          name of the resulting image
+EOF
+	exit 1
+}
+
+annotationsFilePath=""
+checkpointPath=""
+imageName=""
+
+while getopts ":a:c:i:" opt; do
+	case ${opt} in
+	a)
+		annotationsFilePath=$OPTARG
+		;;
+	c)
+		checkpointPath=$OPTARG
+		;;
+	i)
+		imageName=$OPTARG
+		;;
+	:)
+		echo "Option -$OPTARG requires an argument."
+		usage
+		;;
+	\?)
+		echo "Invalid option: -$OPTARG"
+		usage
+		;;
+	esac
+done
+shift $((OPTIND - 1))
+
+if [[ -z $annotationsFilePath || -z $checkpointPath || -z $imageName ]]; then
+	echo "All options (-a, -c, -i) are required."
+	usage
+fi
+
+if ! command -v buildah &>/dev/null; then
+	echo "buildah is not installed. Please install buildah before running 'checkpointctl build' command."
+	exit 1
+fi
+
+if [[ ! -f $annotationsFilePath ]]; then
+	echo "Annotations file not found: $annotationsFilePath"
+	exit 1
+fi
+
+if [[ ! -f $checkpointPath ]]; then
+	echo "Checkpoint file not found: $checkpointPath"
+	exit 1
+fi
+
+newcontainer=$(buildah from scratch)
+
+buildah add "$newcontainer" "$checkpointPath"
+
+while IFS= read -r line; do
+	key=$(echo "$line" | cut -d '=' -f 1)
+	value=$(echo "$line" | cut -d '=' -f 2-)
+	buildah config --annotation "$key=$value" "$newcontainer"
+done <"$annotationsFilePath"
+
+buildah commit "$newcontainer" "$imageName"
+
+buildah rm "$newcontainer"
+
+echo "Checkpoint image created successfully: $imageName"
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,8 +12,10 @@ jobs:
       - uses: actions/checkout@v4
       - name: Install tools
         run: |
-          sudo dnf -y install ShellCheck bats golang criu asciidoctor iptables iproute kmod jq bash bash-completion zsh fish
+          sudo dnf -y install ShellCheck shfmt bats golang criu asciidoctor iptables iproute kmod jq bash bash-completion zsh fish
           sudo modprobe -va ip_tables ip6table_filter nf_conntrack nf_conntrack_netlink
+      - name: Run make shfmt-lint
+        run: make shfmt-lint
       - name: Run make shellcheck
         run: make shellcheck
       - name: Run make all
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,14 @@
 SHELL = /bin/bash
 PREFIX ?= $(DESTDIR)/usr/local
 BINDIR ?= $(PREFIX)/bin
+SCRIPTDIR ?= $(DESTDIR)/usr/libexec
 GO ?= go
 GOPATH := $(shell $(GO) env GOPATH)
 GOBIN := $(shell $(GO) env GOBIN)
 GO_SRC = $(shell find . -name \*.go)
 GO_BUILD = $(GO) build
 NAME = checkpointctl
+SCRIPTNAME = build_image.sh
 
 BASHINSTALLDIR=${PREFIX}/share/bash-completion/completions
 ZSHINSTALLDIR=${PREFIX}/share/zsh/site-functions
@@ -33,17 +35,22 @@ release:
 	CGO_ENABLED=0 $(GO_BUILD) -o $(NAME) -ldflags "-X main.name=$(NAME) -X main.version=${VERSION}"
 
 .PHONY: install
-install: $(NAME) install.completions
+install: $(NAME) install.completions install-scripts
 	@echo "  INSTALL " $<
 	@mkdir -p $(DESTDIR)$(BINDIR)
 	@install -m0755 $< $(DESTDIR)$(BINDIR)
 	@make -C docs install
 
+.PHONY: install-scripts
+install-scripts:
+	@echo "  INSTALL SCRIPTS"
+	@install -m0755 internal/scripts/build_image.sh $(DESTDIR)$(SCRIPTDIR)
+
 .PHONY: uninstall
 uninstall: uninstall.completions
 	@make -C docs uninstall
 	@echo " UNINSTALL" $(NAME)
-	@$(RM) $(addprefix $(DESTDIR)$(BINDIR)/,$(NAME))
+	@$(RM) $(addprefix $(DESTDIR)$(BINDIR)/,$(NAME)) $(addprefix $(DESTDIR)$(SCRIPTDIR)/,$(SCRIPTNAME))
 
 .PHONY: clean
 clean:
@@ -58,9 +65,14 @@ golang-lint:
 .PHONY: shellcheck
 shellcheck:
 	shellcheck test/*bats
+	shellcheck internal/scripts/build_image.sh
+
+.PHONY: shfmt-lint
+shfmt-lint:
+	shfmt -w -d internal/scripts/build_image.sh
 
 .PHONY: lint
-lint: golang-lint shellcheck
+lint: golang-lint shellcheck shfmt-lint
 
 .PHONY: test
 test: $(NAME)
@@ -126,9 +138,10 @@ help:
 	@echo " * completions - generate auto-completion files"
 	@echo " * clean - remove artifacts"
 	@echo " * docs - build man pages"
-	@echo " * lint - verify the source code (shellcheck/golangci-lint)"
+	@echo " * lint - verify the source code (shellcheck/golangci-lint/shfmt-lint)"
 	@echo " * golang-lint - run golangci-lint"
 	@echo " * shellcheck - run shellcheck"
+	@echo " * shfmt-lint - run shfmt on selected shell scripts"
 	@echo " * vendor - update go.mod, go.sum, and vendor directory"
 	@echo " * test - run tests"
 	@echo " * test-junit - run tests and create junit output"
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -31,6 +31,8 @@ func main() {
 
 	rootCommand.AddCommand(cmd.List())
 
+	rootCommand.AddCommand(cmd.BuildCmd())
+
 	rootCommand.Version = version
 
 	if err := rootCommand.Execute(); err != nil {
--- a/cmd/build.go
+++ b/cmd/build.go
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"log"
+
+	"github.com/checkpoint-restore/checkpointctl/internal"
+	"github.com/spf13/cobra"
+)
+
+func BuildCmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "build [checkpoint-path] [image-name]",
+		Short: "Create an OCI image from a container checkpoint archive",
+		RunE:  convertArchive,
+	}
+
+	return cmd
+}
+
+func convertArchive(cmd *cobra.Command, args []string) error {
+	if len(args) != 2 {
+		return fmt.Errorf("please provide both the checkpoint path and the image name")
+	}
+
+	checkpointPath := args[0]
+	imageName := args[1]
+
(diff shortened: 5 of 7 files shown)


Parthiba-Hazra:
I will add test cases gradually.

adrianreber:
Thanks for this PR. It looks like the right approach. Unfortunately, now that the PR exists, I am no longer sure this is right thing to do.

This PR pulls in so many dependencies (over 3000 files), that maybe this was not a good idea after all.

Parthiba-Hazra:
@adrianreber yes I noticed that also, I am not getting any solution of this problem right now, do you have any advise for me ?

rst0git:
> This PR pulls in so many dependencies (over 3000 files), that maybe this was not a good idea after all.

>  I am not getting any solution of this problem right now, do you have any advise for me ?

It is definitely not a good idea to add all these dependencies. A much simpler solution would be to use something like `exec.Command()` and run the `buildah` binary instead.

rst0git:
> This PR introduce a new command `checkpointctl convert`

What about using `checkpointctl build` instead of "convert" as it creates an OCI image similar to `podman build` and `docker build`?

adrianreber:
Not sure about exec. Maybe it should just be a shell script. Not sure.

Parthiba-Hazra:
> Not sure about exec. Maybe it should just be a shell script. Not sure.

If we use shell script, then where should we store it(maybe in $HOME/.checkpointctl/scripts) ? 
and I don't think embedding the script into the go code as string and then writing them out to temporary files or executing them directlty from memory is a good idea.

Parthiba-Hazra:
Now if we use script, we have to ensure that user have buildah installed or we can add buildah installation as a part of checkpointctl installation.

adrianreber:
> Now if we use script, we have to ensure that user have buildah installed or we can add buildah installation as a part of checkpointctl installation.

Yes, we will document it and the actual dependencies is up to the package managers in the distributions. The script can also check for `buildah` and exit with an error message if it is not there.

Parthiba-Hazra:
@adrianreber @rst0git PTAL, If there is no major refactoring required, I can start working on tests.

adrianreber:
Why does the PR now still add over 16000 lines of code? Something is not right.

adrianreber:
Why does the PR now still add over 16000 lines of code? Something is not right.

Parthiba-Hazra:
> Why does the PR now still add over 16000 lines of code? Something is not right.

yes I am checking that.

go.mod
```diff
@@ -11,14 +11,18 @@ require (
...
+	github.com/rivo/uniseg v0.4.4 // indirect
+)
+
 require (
 	github.com/docker/go-units v0.5.0 // indirect
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/klauspost/compress v1.17.7 // indirect
 	github.com/klauspost/pgzip v1.2.6 // indirect
-	github.com/mattn/go-runewidth v0.0.9 // indirect
+	github.com/mattn/go-runewidth v0.0.15 // indirect
```
Parthiba-Hazra, line 24: @adrianreber actually my PR updated this dependency and I think this is required by table writer and thus this dependency pulling the `rivo` .

adrianreber: This should not be part of this PR. Please undo it.

Parthiba-Hazra: yes sure

adrianreber:
Good first try. I must confess I am still unsure which direction we should go, but let's try and see where this leads us.

Please add your script to the `shellcheck` target and please also take a look at the tool `shfmt` and add it during GitHub Actions CI runs. I recently introduce `shfmt` in another project. You can take a look at what I did there: https://github.com/openhpc/ohpc/blob/3.x/tests/ci/Makefile#L74

adrianreber:
For the destination directory of the script I am currently thinking about `/usr/libexec/...`. That seems to be Fedora specific, but the git package for example uses it also extensively for internal scripts on my system:
```shell
$ rpm -ql git | grep libexec
/usr/libexec/git-core/git-contacts
/usr/libexec/git-core/git-credential-netrc
/usr/libexec/git-core/git-filter-branch
/usr/libexec/git-core/git-request-pull
```
We need to figure out what other distributions are using for internal scripts.

internal/scripts/build_image.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+set -euo pipefail
+
```
adrianreber, line 4: bash has a builtin optarg parser which would be nicer than purely positional parameters.

Also add checks that all the input files actually exist.

adrianreber:
I would also say, instead of container engine specific annotations, which are almost not used at all (the only important one if name for CRI-O at least, but we can also change that), let's own the annotations in this package. We should also move the annotation definition to lib, so that all projects reading the annotations can easily import them from checkpointctl. Maybe start almost all annotation with `checkpointctl.annotation`.

@rst0git do you remember which annotation are actually used by Podman? CRI-O is just looking for `io.kubernetes.cri-o.annotations.checkpoint.name`.

rst0git:
We currently use the runtime name to identify an OCI image that contains checkpoint. This functionality needs be extended (in Podman and CRI-O) to verify the comparability of the system before running `criu restore`. For example, we need to make sure that CRIU has [the same version or newer](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L95), the [CPU architecture](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L115) is the same, the [kernel version](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L120) is compatible, etc.

adrianreber:
> We currently use the runtime name to identify an OCI image that contains checkpoint. This functionality needs be extended (in Podman and CRI-O) to verify the comparability of the system before running `criu restore`. For example, we need to make sure that CRIU has [the same version or newer](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L95), the [CPU architecture](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L115) is the same, the [kernel version](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L120) is compatible, etc.

Right. But are any of the annotations already verified today or is that just an idea for the future?

rst0git:
> Right. But are any of the annotations already verified today or is that just an idea for the future?

In Podman we currently use only [`IsCheckpointImage`](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/cmd/podman/utils/utils.go#L116) to check annotations.
```Go
if hostInfo.Host.OCIRuntime.Name != checkpointRuntimeName {
	return false, fmt.Errorf("container image \"%s\" requires runtime: \"%s\"", imgID, checkpointRuntimeName)
}
```

rst0git:
> I would also say, instead of container engine specific annotations, which are almost not used at all (the only important one if name for CRI-O at least, but we can also change that), let's own the annotations in this package. We should also move the annotation definition to lib, so that all projects reading the annotations can easily import them from checkpointctl. Maybe start almost all annotation with checkpointctl.annotation.

This is a good idea.
@Parthiba-Hazra would you like to open a separate pull request that adds the annotations from [Podman](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L67) and [CRI-O](https://github.com/cri-o/cri-o/blob/183f28b7e4ec06b004de4d54b771960e83d2f9fb/pkg/annotations/checkpoint.go#L3)?

Parthiba-Hazra:
> For the destination directory of the script I am currently thinking about `/usr/libexec/...`. That seems to be Fedora specific, but the git package for example uses it also extensively for internal scripts on my system:
> 
> ```shell
> $ rpm -ql git | grep libexec
> /usr/libexec/git-core/git-contacts
> /usr/libexec/git-core/git-credential-netrc
> /usr/libexec/git-core/git-filter-branch
> /usr/libexec/git-core/git-request-pull
> ```
> 
> We need to figure out what other distributions are using for internal scripts.

I thought `/usr/libexec` is debian specific cause somewhere I see that it can be present under `usr/local/libexec`, maybe both possible .
well let me check this.

rst0git:
@Parthiba-Hazra `/usr/libexec` is defined in the [Filesystem Hierarchy Standard (FHS)](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) as follows:

> /usr/libexec includes internal binaries that are not intended to be executed directly by users or shell scripts. Applications may use a single subdirectory under /usr/libexec.

https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch04s07.html

adrianreber:
> > > I would also say, instead of container engine specific annotations, which are almost not used at all (the only important one if name for CRI-O at least, but we can also change that), let's own the annotations in this package. We should also move the annotation definition to lib, so that all projects reading the annotations can easily import them from checkpointctl. Maybe start almost all annotation with checkpointctl.annotation.
> > 
> > 
> > This is a good idea. @Parthiba-Hazra would you like to open a separate pull request that adds the annotations from [Podman](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L67) and [CRI-O](https://github.com/cri-o/cri-o/blob/183f28b7e4ec06b004de4d54b771960e83d2f9fb/pkg/annotations/checkpoint.go#L3)? Once this PR is merged, we can update Podman and CRI-O to use the annotations from checkpointctl.
> 
> @rst0git @adrianreber do we need to keep those annotation's format unchanged or change it to similar what we are doing here(`checkpointctl.annotation`)?

We talked a bit about it and we are not sure. Naming things is hard :wink: .

As this is all about CRIU I think we could go with something like `org.criu.checkpointctl.name`. I would not include `annotation` in the annotation.

Parthiba-Hazra:
> > > > I would also say, instead of container engine specific annotations, which are almost not used at all (the only important one if name for CRI-O at least, but we can also change that), let's own the annotations in this package. We should also move the annotation definition to lib, so that all projects reading the annotations can easily import them from checkpointctl. Maybe start almost all annotation with checkpointctl.annotation.
> > > 
> > > 
> > > This is a good idea. @Parthiba-Hazra would you like to open a separate pull request that adds the annotations from [Podman](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L67) and [CRI-O](https://github.com/cri-o/cri-o/blob/183f28b7e4ec06b004de4d54b771960e83d2f9fb/pkg/annotations/checkpoint.go#L3)? Once this PR is merged, we can update Podman and CRI-O to use the annotations from checkpointctl.
> > 
> > 
> > @rst0git @adrianreber do we need to keep those annotation's format unchanged or change it to similar what we are doing here(`checkpointctl.annotation`)?
> 
> We talked a bit about it and we are not sure. Naming things is hard 😉 .
> 
> As this is all about CRIU I think we could go with something like `org.criu.checkpointctl.name`. I would not include `annotation` in the annotation.

yes, this looks way better and reasonable - `org.criu.checkpointctl.name`.

Parthiba-Hazra:
@adrianreber @rst0git Any update on the annotation's name? Is this the final format - org.criu.checkpointctl.checkpoint.name?

Also, shouldn't this annotation's name be `org.criu.checkpointctl.checkpoint.container` instead of `checkpoint.name`?
```
        // CheckpointAnnotationName is used when creating an OCI image
	// from a checkpoint archive to specify the name of the checkpoint.
	CheckpointAnnotationName = "checkpointctl.annotation.checkpoint.name"
```

rst0git:
> Any update on the annotation's name? Is this the final format - org.criu.checkpointctl.checkpoint.name?

@Parthiba-Hazra If you open another pull request with the updated annotations, we can discuss the format there.

snprajwal:
@Parthiba-Hazra hi! Is there any update on the status of this pull request?

Parthiba-Hazra:
> @Parthiba-Hazra hi! Is there any update on the status of this pull request?

As far as I can remember, this PR can be opened after resolving conflicts as the [annotations](https://github.com/checkpoint-restore/checkpointctl/pull/127) PR was merged.

snprajwal:
@Parthiba-Hazra great! Would you like to rebase this PR and mark it ready for review?

snprajwal:
Superseded by #156

PR #127, opened 2024-04-15
Add annotations for creating OCI images from checkpoint archives

Parthiba-Hazra:
These annotations provide metadata including the container manager, container-name, pod, namespace for creating OCI images from checkpoint archives.


Parthiba-Hazra:
@rst0git @adrianreber Let's first confirm the structure of the annotation. After that, I will add additional annotations for Podman and CRI-O.

adrianreber:
This feels redundant: `checkpointctl.checkpoint`. Let's remove one of those words. I think `checkpointctl` is not really necessary. Even if we said so earlier.

rst0git:
@adrianreber How should we handle backward comparability with CRI-O and Podman? Do we want to deprecate the old annotations and provide support for them up to specified version, or to simply replace them with a "breaking change".

adrianreber:
> @adrianreber How should we handle backward comparability with CRI-O and Podman? Do we want to deprecate the old annotations and provide support for them up to specified version, or to simply replace the annotations with a "breaking change".

For CRI-O it is relatively easy as we only use one annotation. We will just handle both (for some time or forever). For Podman I would personally do the same. No need to break anything for the annotations used by Podman. For the ones not used I would just rename them.

lib/annotations.go
```diff
@@ -0,0 +1,35 @@
+package metadata
+
+const (
+	// CheckpointAnnotationContainerManager is used when creating an OCI image
+	// from a checkpoint archive to specify the name of container manager.
+	CheckpointAnnotationContainerManager = "org.criu.checkpoint.container.manager"
```
rst0git, line 6: The following might be more appropriate: `"org.criu.checkpoint.engine"`

lib/annotations.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// CheckpointAnnotationName is used when creating an OCI image
+	// from a checkpoint archive to specify the name of the checkpoint.
+	CheckpointAnnotationName = "org.criu.checkpoint.container"
```
rst0git, line 10: "container" could mean many different things.
What about `"org.criu.checkpoint.name"`?

lib/annotations.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// CheckpointAnnotationPod is used when creating an OCI image
+	// from a checkpoint archive to specify the name of the pod associated with the checkpoint.
+	CheckpointAnnotationPod = "org.criu.checkpoint.pod"
+
+	// CheckpointAnnotationNamespace is used when creating an OCI image
+	// from a checkpoint archive to specify the namespace of the pod associated with the checkpoint.
+	CheckpointAnnotationNamespace = "org.criu.checkpoint.namespace"
```
rst0git, line 18: `"org.criu.checkpoint.pod.namespace"`

lib/annotations.go
```diff
@@ -0,0 +1,35 @@
...
+	// from a checkpoint archive to specify the ID of the root filesystem image associated with the checkpoint.
+	CheckpointAnnotationRootfsImageID = "org.criu.checkpoint.rootfsImageID"
+
+	// CheckpointAnnotationRootfsImageName is used when creating an OCI image
+	// from a checkpoint archive to specify the name of the root filesystem image associated with the checkpoint.
+	CheckpointAnnotationRootfsImageName = "org.criu.checkpoint.rootfsImageName"
+
+	// CheckpointAnnotationRuntimeName is used when creating an OCI image
+	// from a checkpoint archive to specify the runtime used on the host where the checkpoint was created.
+	CheckpointAnnotationRuntimeName = "org.criu.checkpoint.runtime.name"
```
rst0git, line 34: It would be nice to also add `"org.criu.checkpoint.runtime.version"`

lib/annotations.go
```diff
@@ -0,0 +1,35 @@
...
+	// from a checkpoint archive to specify the name of container manager.
+	CheckpointAnnotationContainerManager = "org.criu.checkpoint.container.manager"
+
+	// CheckpointAnnotationName is used when creating an OCI image
+	// from a checkpoint archive to specify the name of the checkpoint.
+	CheckpointAnnotationName = "org.criu.checkpoint.container"
+
+	// CheckpointAnnotationPod is used when creating an OCI image
+	// from a checkpoint archive to specify the name of the pod associated with the checkpoint.
+	CheckpointAnnotationPod = "org.criu.checkpoint.pod"
```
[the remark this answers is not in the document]
adrianreber: Kubernetes also has the concept of pod UID and there is pod ID in the container engine.

lib/annotations.go
adrianreber: `pod.namespace` seems strange to me. Because the pod is in a namespace. Should we do it hierarchically?

- namespace.name
- namespace.pod.name
- namespace.pod.uid
- namespace.pod.id
- namespace.pod.container.id
- namespace.pod.container.name

But that seems to complicated.

lib/annotations.go
Parthiba-Hazra: Yes you are right

lib/annotations.go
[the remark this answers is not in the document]
rst0git: I'm not sure if saving the pod UID / ID with annotations would be useful. These are automatically generated and used only during runtime: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids

lib/annotations.go
rst0git: According to the [rules for annotations](https://github.com/opencontainers/image-spec/blob/main/annotations.md#rules), it is considered best practice to "namespace the keys":

> - Annotations MUST be a key-value map where both the key and value MUST be strings.
> - While the value MUST be present, it MAY be an empty string.
> - Keys MUST be unique within this map, and .
> - Keys SHOULD be named using a reverse domain notation - e.g. `com.example.myKey`.
> - The prefix `org.opencontainers` is reserved for keys defined in Open Container Initiative (OCI) specifications and MUST NOT be used by other specifications and extensions.
> - Keys using the `org.opencontainers.image` namespace are reserved for use in the OCI Image Specification and MUST NOT be used by other specifications and extensions, including other OCI specifications.
> - If there are no annotations then this property MUST either be absent or be an empty map.
> - Consumers MUST NOT generate an error if they encounter an unknown annotation key.

lib/annotations.go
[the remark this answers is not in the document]
adrianreber: pod UID is important to have for restore. If we need to replace the old pod UID with the new pod UID it is helpful to easily get the old pod UID. Currently I have to look into the `spec.dump` and extract it from there. If we have one place where it is stored for easy access. Makes sense from my point of view.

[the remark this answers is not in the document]
rst0git: We also need to keep in mind the fast startup use-case, where a single checkpoint is used to start multiple replicas. This is similar to `podman container restore --name <container name>`, that allows us to restore multiple containers from a single checkpoint with different names.

lib/annotations.go
Parthiba-Hazra: @rst0git Won't it be misunderstood as the name of the checkpoint? Here, this annotation is used to specify the container name.
should we go with - `org.criu.checkpoint.container.name`

rst0git: > should we go with - org.criu.checkpoint.container.name

Sounds good

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// CheckpointAnnotationDistributionName is used by Container Checkpoint when
+	// creating a checkpoint image to specify the name of host distribution on
+	// which the checkpoint was created.
+	CheckpointAnnotationDistributionName = "org.criu.checkpoint.distribution.name"
+
+	// CheckpointAnnotationCRIOVersion is used by Container Checkpoint when
+	// creating a checkpoint image to specify the version of CRI-O used on the
+	// host where the checkpoint was created.
+	CheckpointAnnotationCRIOVersion = "org.criu.checkpoint.cri-o.version"
```
rst0git, line 104: This is not necessary because we have `"org.criu.checkpoint.engine.version"`

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// CheckpointAnnotationDistributionVersion is used by Container Checkpoint
+	// when creating a checkpoint image to specify the version of host
+	// distribution on which the checkpoint was created.
+	CheckpointAnnotationDistributionVersion = "org.criu.checkpoint.distribution.version"
+
+	// CheckpointAnnotationDistributionName is used by Container Checkpoint when
+	// creating a checkpoint image to specify the name of host distribution on
+	// which the checkpoint was created.
+	CheckpointAnnotationDistributionName = "org.criu.checkpoint.distribution.name"
```
rst0git, line 99: I'm not sure if distribution name and version are relevant for checkpoint/restore.

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// CheckpointAnnotationRawImageName is used by Container Checkpoint when
+	// creating a checkpoint image to specify the original unprocessed name of
+	// the image used to create the container (as specified by the user).
+	CheckpointAnnotationRawImageName = "org.criu.checkpoint.rawImageName"
+
+	// CheckpointAnnotationPodmanVersion is used by Container Checkpoint when
+	// creating a checkpoint image to specify the version of Podman used on the
+	// host where the checkpoint was created.
+	CheckpointAnnotationPodmanVersion = "org.criu.checkpoint.podman.version"
```
rst0git, line 64: This is not necessary because we have "org.criu.checkpoint.engine.version"

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// CheckpointAnnotationRuntimeVersion is used by Container Checkpoint when
+	// creating a checkpoint image to specify the version of runtime used on the
+	// host where the checkpoint was created.
+	CheckpointAnnotationRuntimeVersion = "org.criu.checkpoint.runtime.version"
+
+	// CheckpointAnnotationRawImageName is used by Container Checkpoint when
+	// creating a checkpoint image to specify the original unprocessed name of
+	// the image used to create the container (as specified by the user).
+	CheckpointAnnotationRawImageName = "org.criu.checkpoint.rawImageName"
```
rst0git, line 59: Would you be able to move this to be just after `rootfsImageName`?

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+
+	// CheckpointAnnotationNamespace is used by Container Checkpoint when
+	// creating a checkpoint image to specify the namespace of the pod associated
+	// with the checkpoint.
+	CheckpointAnnotationNamespace = "org.criu.checkpoint.pod.namespace"
+
+	// CheckpointAnnotationRootfsImage is used by Container Checkpoint when
+	// creating a checkpoint image to specify the root filesystem image associated
+	// with the checkpoint.
+	CheckpointAnnotationRootfsImage = "org.criu.checkpoint.rootfsImage"
```
rst0git, line 34: What is the purpose of `org.criu.checkpoint.rootfsImage` and how is it different from `org.criu.checkpoint.rootfsImageName`?

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
+package metadata
+
+const (
+	// CheckpointAnnotationEngine is used by Container Checkpoint when
+	// creating a checkpoint image to specify the name of the container engine.
+	CheckpointAnnotationEngine = "org.criu.checkpoint.engine.name"
+
+	// CheckpointAnnotationEngineVersion is used by Container Checkpoint when
+	// creating a checkpoint image to specify the version of the container engine.
```
rst0git, line 9:
```
// CheckpointAnnotationEngineVersion specifies the version of the container engine used to create the checkpoint.
```

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+	CheckpointAnnotationEngineVersion = "org.criu.checkpoint.engine.version"
+
+	// CheckpointAnnotationName is used by Container Checkpoint when
+	// creating a checkpoint image to specify the name of the checkpoint.
```
rst0git, line 13:
```
// CheckpointAnnotationName specifies the original human-readable name for the container.
```

lib/annotations.go
```diff
@@ -0,0 +1,105 @@
...
+	CheckpointAnnotationName = "org.criu.checkpoint.name"
+
+	// CheckpointAnnotationPod is used by Container Checkpoint when
+	// creating a checkpoint image to specify the name of the pod associated
+	// with the checkpoint.
```
rst0git, line 18:
```
// CheckpointAnnotationPod specifies the original Pod name associated with the checkpointed container.
```

lib/annotations.go
Parthiba-Hazra: Actually, I was adding the annotation from podman and CRI-O and keep it unchanged as, I am not aware of specific use cases for each annotations. And if this is not relevant then we can remove this.

lib/annotations.go
Parthiba-Hazra: I was copying this type of annotation from the [podman](https://github.com/containers/podman/blob/447d3e2a44922b83495a331528bbd8922277e067/libpod/define/annotations.go#L67) and [CRI-O](https://github.com/cri-o/cri-o/blob/183f28b7e4ec06b004de4d54b771960e83d2f9fb/pkg/annotations/checkpoint.go#L3) as those might be used for creating checkpoint image in that particular engine.

lib/annotations.go
Parthiba-Hazra: Example -
```
"checkpointctl.annotation.checkpoint.rootfsImage": "docker.io/library/nginx@sha256:52478f8cd6a142fd462f0a7614a7bb064e969a4c083648235d6943c786df8cc7",
               "checkpointctl.annotation.checkpoint.rootfsImageID": "92b11f67642b62bbb98e7e49169c346b30e20cd3c1c034d31087e46924b9312e",
               "checkpointctl.annotation.checkpoint.rootfsImageName": "docker.io/library/nginx:latest",
```
I think `org.criu.checkpoint.rootfsImage`  containing some kind of hashed value in it, maybe calculated by some kind of cryptographic function like SHA256 for uniquely identify an image.

rst0git: @adrianreber I noticed that you have introduced `rootfsImage` (https://github.com/cri-o/cri-o/commit/38774e14a4f9e552cb01ccf71544eed8039e1da6) to make sure that we pull exactly the same base image during restore, and `rootfsImageName` seems to be redundant/irrelevant (https://github.com/cri-o/cri-o/commit/1eae23ee64548a90642e605ba920946970a71e2d). Would it make sense to remove `rootfsImageName` from the annotations here?

lib/annotations.go
adrianreber: I agree. If possible there should not be any container engine specific annotation.

lib/annotations.go
adrianreber: As mentioned above, I do not think we need any container engine specific annotations.

lib/annotations.go
adrianreber: The whole base rootfs image is currently not clear in different container engines from the checkpoint view.

We basically need three things. The image name the user has specified, the image name (including sha) which is actually used and probably the sha of the current image. In CRI-O this information is currently difficult to get, but should soon be available from what I have been told. So we need something like this:

- rootfsImageName (registry/path/image@sha:123456)
- rootfsImageUserRequested (registry/path/image:tag)
- rootfsImageSha or rootfsImageRef (123456)

The last one is not 100% necessary, but instead of parsing rootfsImageName to extract the sha it is easier to have it additionally listed.

Parthiba-Hazra: > rootfsImageUserRequested (registry/path/image:tag)

@adrianreber I don't understand the significance of the term 'UserRequested' here. Could you break it down for me a little?

rst0git: There is a function in CRI-O called [UserRequestedImage](https://github.com/cri-o/cri-o/blob/63cfc3b92ce26dfc6bc068529065d3c0687ce827/internal/oci/container.go#L360C21-L360C39):

>  UserRequestedImage returns the users' input originally used to find imageID; it might evaluate to a different image (or to a different kind of reference!) at any future time.

lib/annotations.go
rst0git: I believe we added these annotations because some distributions, like Ubuntu, include non-upstream kernel patches that can cause CRIU to fail:
- https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1857257
- https://bugs.launchpad.net/ubuntu/+source/linux/+bug/1967924

rst0git: @adrianreber What do you think?

rst0git: @Parthiba-Hazra Would you be able to add a comment that clarifies the reason for using these annotations?
e.g., some distributions may include non-upstream patches that can cause CRIU to fail.

Parthiba-Hazra: sure :+1:

rst0git (approved):

rst0git:
@adrianreber PTAL

lib/annotations.go
```diff
@@ -0,0 +1,65 @@
...
+	CheckpointAnnotationConmonVersion = "org.criu.checkpoint.conmon.version"
+
+	// CheckpointAnnotationHostArch specifies the CPU architecture of the host where the checkpoint was created.
+	CheckpointAnnotationHostArch = "org.criu.checkpoint.host.arch"
+
+	// CheckpointAnnotationHostKernel specifies the kernel version used by the host where the checkpoint was created.
+	CheckpointAnnotationHostKernel = "org.criu.checkpoint.host.kernel"
+
+	// CheckpointAnnotationCgroupVersion specifies the cgroup version used by the host where the checkpoint was created.
+	CheckpointAnnotationCgroupVersion = "org.criu.checkpoint.cgroups.version"
```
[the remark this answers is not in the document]
snprajwal: Ah no, I meant that we should either use singular in both the variable name and string, or plural. Either one works.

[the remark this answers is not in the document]
adrianreber: Right, I have not seen that. Singular feels right to me.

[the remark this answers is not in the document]
adrianreber: @Parthiba-Hazra please make that change and this looks ready to go.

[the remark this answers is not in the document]
Parthiba-Hazra: Done :+1:

adrianreber:
The test failure is unexpected. Seems like some project changed the minimum Go version without changing its version. Strange.

I guess we need another PR doing `make vendor` before we can merge this one.

Parthiba-Hazra:
> The test failure is unexpected. Seems like some project changed the minimum Go version without changing its version. Strange.
> 
> I guess we need another PR doing `make vendor` before we can merge this one.

@adrianreber I opened one here #133

adrianreber:
Please rebase.

PR #135, opened 2024-06-17
Additional container metadata from containerd

adrianreber:
containerd structures the internal metadata differently. For a restore the status.dump file is also necessary.

### DIFF
--- a/lib/metadata.go
+++ b/lib/metadata.go
@@ -16,6 +16,7 @@ const (
 	// container archive
 	ConfigDumpFile             = "config.dump"
 	SpecDumpFile               = "spec.dump"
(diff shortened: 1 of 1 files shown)


lib/metadata.go
```diff
@@ -16,6 +16,7 @@ const (
 	// container archive
 	ConfigDumpFile             = "config.dump"
 	SpecDumpFile               = "spec.dump"
+	StatusDumpFile             = "status.dump"
```
rst0git, line 19: What is the content/purpose of this file? How is it going to be used during restore?

adrianreber: I updated the commit message.

rst0git (approved):
LGTM

PR #140, opened 2024-07-19
ci: update codecov-action from v3 to v4

rst0git:


snprajwal:
@rst0git the coverage workflow on the main branch succeeds as of now without any problems: https://github.com/checkpoint-restore/checkpointctl/actions/runs/10009967467/job/27670052231

rst0git:
> the coverage workflow on the main branch succeeds as of now without any problems: https://github.com/checkpoint-restore/checkpointctl/actions/runs/10009967467/job/27670052231

@snprajwal The workflow fails is unable to access the upload token in pull requests (e.g., https://github.com/checkpoint-restore/checkpointctl/pull/141, https://github.com/checkpoint-restore/checkpointctl/pull/140)

snprajwal:
Right, so it seems GitHub does not allow accessing secrets implicitly in workflows triggered on commits outside the upstream repo: https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

Adding a `with` item with the secret should fix this for us?

snprajwal:
Going by the docs, the `pull_request` trigger does not allow access to secrets, but the `pull_request_target` trigger does. Can we try this?

rst0git:
@snprajwal It looks like tokenless uploading works with PRs:
- https://github.com/codecov/codecov-action/issues/1431#issuecomment-2150431225
- https://github.com/netobserv/network-observability-console-plugin/pull/531

.github/workflows/coverage.yml
```diff
@@ -23,9 +22,8 @@ jobs:
         run: sudo -E make coverage
 
       - name: Upload coverage reports to Codecov
-        uses: codecov/codecov-action@v3
+        uses: codecov/codecov-action@v4
         with:
-          fail_ci_if_error: true
+          fail_ci_if_error: false
```
snprajwal, line 27: It might be good to retain this in order to watch for failures unrelated to the token error. As far as I can see, it shouldn't affect the job status for tokenless upload if everything goes right.

snprajwal (approved):

snprajwal:
@rst0git by the way, reading the issue you linked, it looks like the token is required for PRs coming from the base repo + commits on the base repo itself, and forks can use the tokenless version. If that is indeed the case, we might require another check to decide if the token should be used or not (based on whether the trigger is `push` or `pull_request`)

rst0git:
@snprajwal This problem seems to be fixed with https://github.com/codecov/codecov-action/pull/1437.

PR #144, opened 2024-10-15
Fix location of Open Files in tree

adrianreber:
Without this patch the output used to look something like this
```
│   └── [1]  bash
│       ├── Environment variables
│       ├── [8]  /usr/bin/coreutils
│       │   ├── Environment variables
│       │   └── Open files
│       │       └── [root]  /
│       ├── [7]  /usr/bin/python3
│       │   ├── Environment variables
│       │   ├── Open files
│       │   │   └── [root]  /
│       │   └── Open sockets
│       │       └── [TCP (LISTEN)]  0.0.0.0:8088 -> 0.0.0.0:0 (↑ 16.0 KB ↓ 128.0 KB)
│       └── Open files
│           └── [root]  /
```
Which is wrong as the list of open files (and sockets) for the first
process is after the child processes. With this change the order is now
correct:
```
│   └── [1]  bash
│       ├── Environment variables
│       ├── Open files
│       │   └── [root]  /
│       ├── [7]  /usr/bin/python3
│       │   ├── Environment variables
│       │   ├── Open files
│       │   │   └── [root]  /
│       │   └── Open sockets
│       │       └── [TCP (LISTEN)]  0.0.0.0:8088 -> 0.0.0.0:0 (↑ 16.0 KB ↓ 128.0 KB)
│       └── [8]  /usr/bin/coreutils
│           ├── Environment variables
│           └── Open files
│               └── [root]  /
```
Each process has its own "Open files" and "Open sockets". It was always
correct for "Environment variables".

### DIFF
--- a/internal/tree.go
+++ b/internal/tree.go
@@ -44,29 +44,39 @@ func RenderTreeView(tasks []Task) error {
 				}
 			}
 
-			if err = addPsTreeToTree(tree, psTree, task.OutputDir); err != nil {
+			fds, err := func() ([]*crit.Fd, error) {
+				if !Files {
+					return nil, nil
+				}
+				c := crit.New(nil, nil, filepath.Join(task.OutputDir, "checkpoint"), false, false)
+				fds, err := c.ExploreFds()
+				if err != nil {
+					return nil, fmt.Errorf("failed to get file descriptors: %w", err)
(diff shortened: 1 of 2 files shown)


rst0git (approved):
LGTM

PR #149, opened 2024-10-23
Recursively add metadata JSON fields to the tree

rst0git:
Some annotations contain a JSON string, and this string may include nested arrays and maps. This pull request updates `addPodInfoToTree()` to recursively add these fields to the output tree using pretty print.

Before:

        ├── kubectl.kubernetes.io/last-applied-configuration
        │   ├── metadata: map[annotations:map[] name:counter-pod namespace:default]
        │   ├── spec: map[containers:[map[args:[-c i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done] command:[sh] image:busybox:latest name:counter-container]]]

After:

        ├── kubectl.kubernetes.io/last-applied-configuration
        │   ├── apiVersion: v1
        │   ├── kind: Pod
        │   ├── metadata:
        │   │   ├── annotations:
        │   │   ├── name: counter-pod
        │   │   └── namespace: default
        │   └── spec:
        │       └── containers: 
        │           └── [0]:
        │               ├── args: 
        │               │   ├── [0]: -c
        │               │   └── [1]: i=0; while true; do echo $i; i=$(expr $i + 1); sleep 1; done
        │               ├── command: 
        │               │   └── [0]: sh
        │               ├── image: busybox:latest
        │               └── name: counter-container

### DIFF
--- a/internal/tree.go
+++ b/internal/tree.go
@@ -335,9 +335,8 @@ func addPodInfoToTree(tree treeprint.Tree, info *checkpointInfo) {
 					continue
 				}
 				localTree := annotationTree.AddBranch(key)
-				for labelKey := range local {
-					localTree.AddBranch(fmt.Sprintf("%s: %s", labelKey, local[labelKey]))
-				}
+				// Recursively add array/map JSON fields to the tree
+				addMapToTree(localTree, local)
 			case "io.kubernetes.cri-o.Volumes":
 				// We know that some annotations contain a JSON string we can pretty print
 				var local []mountAnnotations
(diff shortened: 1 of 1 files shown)


adrianreber:
Nice, I was also thinking about something like that. I am not sure about:
```
├── args: [
...
└── ]
```
That look a bit strange. Could this just be left away? Or why do you print it? Any special reason it might be useful?

snprajwal:
+1, I think we can just leave the braces out, similar to `metadata: `

rst0git:
Thanks, I've updated the pull request to remove the brackets in the output.

snprajwal (approved):

PR #156, opened 2025-03-04
Add build command for OCI images

rst0git:
This pull request contains a revised version of the changes in #125, adding a `checkpointctl build` command that allows for building OCI images from checkpoint archives.

Example:

```
checkpointctl build ./checkpoint.tar quay.io/foo/bar:latest
buildah push quay.io/foo/bar:latest
```

### DIFF
--- a/checkpointctl.go
+++ b/checkpointctl.go
@@ -31,6 +31,8 @@ func main() {
 
 	rootCommand.AddCommand(cmd.List())
 
+	rootCommand.AddCommand(cmd.BuildCmd())
+
 	rootCommand.Version = version
 
 	if err := rootCommand.Execute(); err != nil {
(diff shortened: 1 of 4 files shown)


internal/image_from_checkpoint_archive.go
```diff
@@ -0,0 +1,142 @@
...
+	"log"
+	"os"
+	"os/exec"
+
+	metadata "github.com/checkpoint-restore/checkpointctl/lib"
+)
+
+const (
+	BUILD_SCRIPT  = "/usr/libexec/build_image.sh"
+	PODMAN_ENGINE = "libpod"
```
behouba, line 17: I am curious about where we will need these exported Constants. Maybe PODMAN_ENGIME could be used [here](https://github.com/checkpoint-restore/checkpointctl/blob/15dd6b5ae2e32f3435650f15826dd367a70fc984/internal/container.go#L184). But what about BUILD_SCRIPT?

snprajwal (approved):

behouba (approved):
LGTM

PR #157, opened 2025-03-04
ci: run as root within container

rst0git:
When running with default user, it fails with the following error:

```
Run sudo dnf -y install ShellCheck bats golang criu asciidoctor iptables iproute kmod jq bash bash-completion zsh fish
sudo: PAM account management error: Authentication service cannot retrieve authentication info
sudo: a password is required
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -7,21 +7,21 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: registry.fedoraproject.org/fedora:latest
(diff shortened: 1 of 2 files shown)


snprajwal:
@rst0git the lint job is failing because of an update to the golangci-lint config spec. `deadline` is now called `timeout` - we would have to make that change in our `.golangci-lint.yml`.

snprajwal:
#158 should fix this, PTAL

PR #158, opened 2025-03-04
fix(ci): move from Fedora to Ubuntu

snprajwal:
Previously, we were using Fedora as Ubuntu did not contain the updated
binaries for bats. CRIU was also outdated on the main repositories.
Due to a recent change in the way GitHub runners handle privileged
containers, the old Fedora setup no longer works. This is the easiest
solution that allows us to retain sudo-based local tests that can also
be reused on CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,14 +5,18 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
-    container:
-      image: registry.fedoraproject.org/fedora:latest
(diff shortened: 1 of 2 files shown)


rst0git (approved):

adrianreber:
What are the different package versions in Ubuntu. It usually is much older and not updated.

snprajwal:
> What are the different package versions in Ubuntu. It usually is much older and not updated.

```
bats=1.10.0-1
shellcheck=0.9.0
```
Latest bats is v1.11, and shellcheck is v0.10.0. They're lagging by one version, I think it should be fine since we don't depend on any bleeding edge feature.

PR #167, opened 2025-08-01
Replace olekukonko/tablewriter with text/tabwriter

adrianreber:
After some anonymous third party review, I completed changed the approach to this PR. Instead of using the external dependency tablewriter this now uses text/tabwriter. This reduces included code.

Drop external tablewriter dependency in favor of Go's built-in text/tabwriter package to reduce dependencies and simplify maintenance.

Changes:
- Replace tablewriter.NewWriter() with tabwriter.NewWriter() in all table  display functions
- Update table formatting logic to use tab-separated output with headers  and separator lines
- Remove olekukonko/tablewriter and related dependencies from go.mod
- Update test expectations to match new table output format
- Fix test line number references after table format changes

All tests pass with the new implementation.

Assisted-by: Claude AI for dependency replacement and test updates


adrianreber:
@rst0git @snprajwal This update was created with Claude. The main reason to create was that dependabot now combines all changes and the tablewriter change breaks API. That is why I wanted to do it in a separate PR.

The new tablewriter comes with new features and new dependencies which increase the binary size. Not sure if this is too much. If I am able to correctly read the number it is 700KB larger than before.

adrianreber:
Let me also update README.md.

adrianreber:
I told claude to update the PR with the changes @rst0git suggested.

PR #180, opened 2025-12-16
workflows: add PR comment on binary size check failure

adrianreber:
Document the existing bloat-ok label functionality in README.md for bypassing the binary size check. Add support for posting a PR comment when the size check fails, which helps contributors from forks understand why the check failed and how to resolve it. This uses a two-workflow approach: the main workflow uploads artifacts, and a separate workflow triggered by workflow_run posts the comment with proper permissions.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/check-size-comment.yml
+++ b/.github/workflows/check-size-comment.yml
@@ -1,0 +1,95 @@
+name: Check binary size comment
+
+on:
+  workflow_run:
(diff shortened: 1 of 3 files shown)


adrianreber:
Tests are currently broken because of https://github.com/checkpoint-restore/criu/issues/2847

#179 should fix that

PR #181, opened 2025-12-16
Docs: document asciidoctor and Go version requirements and ignore install artifacts

Lorygold:
This PR makes two small but important updates:

* **README** 
  * Added **asciidoctor** to the list of required packages. Without it, make install fails during documentation generation.
  * docs: documented Go version issues on Debian-based distributions

* **.gitignore** – Added the `install/` folder to prevent generated files (e.g., completions) from being tracked by Git.

These changes improve the build/install process and keep the repository clean from artifacts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,6 +205,10 @@ then uploads the newly created OCI image to the container registry, making it av
     ```console
     make
     ```
+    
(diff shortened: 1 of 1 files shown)


adrianreber:
Please use Signed-off-by for your commit.

You wrote a really good PR description please also add that as commit message.

PR #183, opened 2025-12-18
docs: document asciidoctor and Go version requirements and ignore install artifacts

Lorygold:
This PR makes two small but important updates:

* **README** 
  * Added **asciidoctor** to the list of required packages. Without it, make install fails during documentation generation.
  * docs: documented Go version issues

* **.gitignore** – Added the `install/` folder to prevent generated files (e.g., completions) from being tracked by Git.

These changes improve the build/install process and keep the repository clean from artifacts.

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,14 +191,16 @@ then uploads the newly created OCI image to the container registry, making it av
     On Fedora, CentOS and related distributions:
 
     ```console
-    sudo yum install -y go make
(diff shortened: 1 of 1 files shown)


adrianreber:
@rst0git Not sure I agree with your latest change dropping the golang from the package manager line. If it works, no need to install a package directly.

PR #185, opened 2026-01-11
feat: add support for checkpoints with tasks in non-alive state

### LINKED ISSUE #184: Add support for checkpoints with tasks in non-alive state
CRIU includes the state of all processes in a container, including both alive, stopped, and zombie processes:

```C
/*
 * Task state, as returned by compel_wait_task()
 * and used in arguments to compel_resume_task().
 */
enum __compel_task_state {
	COMPEL_TASK_ALIVE = 0x01,
	COMPEL_TASK_DEAD = 0x02,
	COMPEL_TASK_STOPPED = 0x03,
	COMPEL_TASK_ZOMBIE = 0x06,
	/* Don't ever change the above values, they are used by CRIU! */

	COMPEL_TASK_MAX = 0x7f
};
```
However, checkpointctl currently does not take into account the state of a process when parsing a checkpoint archive. This results in errors like the following:

```console
$ crit show core-117.img 
{
    "magic": "CORE",
    "entries": [
        {
            "mtype": "X86_64",
            "tc": {
                "task_state": 2,
                "exit_code": 0,
                "personality": 0,
                "flags": 0,
                "blk_sigset": "0x0",
                "comm": "sleep"
            }
        }
    ]
}

$  checkpointctl inspect --ps-tree-env test.tar
Error: failed to get process tree: error opening binary file: open /tmp/user/0/checkpointctl3450662070/checkpoint/pagemap-117.img: no such file or directory
```

Since zombie processes don't use memory, CRIU does not create pagemap and pages images. We need to check the "status" of tasks before opening these image files.

TusharMohapatra07:
This PR:

- Adds state constants and checks if a process has memory pages. determines if a process should have memory pages based on it's state.
- Modifies memory access functions to check the task state before attempting to open memory-related image files. For zombie/dead processes, these functions returns nil values instead of attempting to open non-existent files.
- updates calls in the codebase to retrieve task state from the core data and passes it to the memory access functions.
- Adds suitable output on user-facing side

fixes #184

### DIFF
--- a/test/checkpointctl.bats
+++ b/test/checkpointctl.bats
@@ -681,7 +681,7 @@ function teardown() {
 	( cd "$TEST_TMP_DIR1" && tar cf "$TEST_TMP_DIR2"/test.tar . )
 	checkpointctl memparse --search=PATH --context=10 "$TEST_TMP_DIR2"/test.tar --pid=1
 	[ "$status" -eq 0 ]
-	[[ ${lines[3]} == *"PATH"* ]]
+	[[ ${lines[2]} == *"PATH"* ]]
 }
 
 @test "Run checkpointctl memparse with --search-regex='HOME=([^?]+)' " {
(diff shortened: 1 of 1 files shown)


test/checkpointctl.bats
```diff
@@ -713,6 +713,52 @@ function teardown() {
 	[[ ${lines[0]} == *"no process with PID 9999"* ]]
 }
 
+@test "Run checkpointctl memparse handles processes without pagemap files" {
```
adrianreber: But the pagemap files seems to be included. Not sure what this test is supposed to do.

TusharMohapatra07:
Hey @adrianreber! To test actual zombie behaviour, I think we need to modify/create binary test data for it. Am i right ?

TusharMohapatra07:
Hey @adrianreber! Modified the tests a bit. Used crit encode and decode to set the task state to that of zombie/dead (for simulation). For tests that do need memory files, i've added comments explaining why we need them so. Please take a look and let me know if i can improve the testing in any other way :)

rst0git:
@TusharMohapatra07 I have a few suggestions and went ahead and revised your patch. I will share the updated version later today.

TusharMohapatra07:
Sure! Thanks

PR #186, opened 2026-01-12
Upgrade golangci-lint to v2

o-love:
Modify golangci lint config with breaking changes from v1 to v2.

Done using `golangci-lint migrate` tool following https://golangci-lint.run/docs/product/migration-guide/


adrianreber:
We are also calling the golangci-lint binary in the Makefile, do we need to change anything there? I guess the binary name has not changed, right? Is there a minimum version of golangci-lint necessary to understand the new configuration file?

PR #199, opened 2026-02-13
lib: Add CheckpointOptions struct and parsing for CRI options

adrianreber:
Add support for parsing checkpoint options from the Kubernetes CRI CheckpointPodRequest map[string]string options field. The new ParseCheckpointOptions function validates known options and their expected types (bool, int, string).

Known options: leave-running, tcp-established, ghost-limit, network-lock

Boolean values accept: yes, no, true, false, on, off, 0, 1

Generated with Claude Code (https://claude.ai/code)

### DIFF
--- a/lib/checkpoint_options.go
+++ b/lib/checkpoint_options.go
@@ -1,0 +1,136 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package metadata
+
+import (
+	"fmt"
+)
+
+// OptionType represents the type of a checkpoint option value
+type OptionType int
+
+const (
+	OptionTypeBool OptionType = iota
+	OptionTypeInt
+	OptionTypeString
+)
+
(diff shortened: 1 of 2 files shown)


lib/checkpoint_options.go
```diff
@@ -0,0 +1,136 @@
...
+		Help: "leave container(s) in running state after checkpointing",
+	},
+}
+
+// CheckpointOptions represents the options from a CheckpointPodRequest
+type CheckpointOptions struct {
+	// LeaveRunning leaves the processes running in the container after checkpointing
+	LeaveRunning bool `json:"leaveRunning,omitempty"`
+	// TCPEstablished enables support for established TCP connections in the checkpoint
+	TCPEstablished bool `json:"tcpEstablished,omitempty"`
```
rst0git: We should also add tcp-close.

PR #205, opened 2026-03-01
feat: add edit subcommand

### LINKED ISSUE #169: Remapping of TCP listen ports
In many scenarios, users need to change the TCP port bound to a socket before restoring a container (e.g., https://github.com/checkpoint-restore/criu/issues/2710). CRIU saves this port in `files.img`, which is included in the checkpoint tar archive.

Currently, the only way to restore a container with a different port is to manually edit the checkpoint before restore. It would be great if `checkpointctl` provides such functionality, similar to the [`--tcp-listen-remap`](https://github.com/checkpoint-restore/criu-image-streamer/pull/7) option of criu-image-streamer or the [edit_files_img.py](https://github.com/stano45/p4containerflow/blob/main/scripts/edit_files_img.py) script.

ImranullahKhann:
Currently only supports tcp listen port remapping. Made the following changes:
- Added the edit command in cmd/.
- Added edit_archive.go in internals/ for the main edit logic.
- Added archive_modifiers.go inside internals/. It contains the modifier logic for any entry that would need changes in the archive. 
- Added unit and integration tests for verifying the functionality.

Kept extensibility in mind, for any additional edit logic in the future, we can reuse the same function tarStreamRewrite (in edit_archive.go) and write additional modifier functions in archive_modifiers.go.

Closes https://github.com/checkpoint-restore/checkpointctl/issues/169

### DIFF
--- a/internal/archive_modifiers.go
+++ b/internal/archive_modifiers.go
@@ -1,0 +1,237 @@
+// SPDX-License-Identifier: Apache-2.0
+
+package internal
+
+import (
+	"archive/tar"
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+	"strconv"
+
+	"github.com/checkpoint-restore/go-criu/v8/crit"
+	"github.com/checkpoint-restore/go-criu/v8/crit/images/fdinfo"
+)
+
+// TCP_LISTEN state value from the Linux kernel
+const tcpListenState = 10
+
+// remapFilesImg decodes a CRIU files.img binary image, remaps the source port
+// of any TCP listen socket matching oldPort to newPort, and re-encodes the image.
+func remapFilesImg(hdr *tar.Header, content io.Reader, oldPort, newPort uint32) (*tar.Header, []byte, error) {
+	// crit.New requires *os.File, so write the tar entry content to a temp file
+	tmpIn, err := os.CreateTemp("", "files-img-in-*.img")
+	if err != nil {
+		return nil, nil, fmt.Errorf("creating temp input file: %w", err)
+	}
+	defer os.Remove(tmpIn.Name())
+	defer tmpIn.Close()
+
+	if _, err := io.Copy(tmpIn, content); err != nil {
+		return nil, nil, fmt.Errorf("writing to temp file: %w", err)
+	}
+	if _, err := tmpIn.Seek(0, 0); err != nil {
+		return nil, nil, fmt.Errorf("seeking temp file: %w", err)
+	}
+
+	// Decode the binary image
+	c := crit.New(tmpIn, nil, "", false, false)
+	img, err := c.Decode(&fdinfo.FileEntry{})
+	if err != nil {
+		return nil, nil, fmt.Errorf("decoding files.img: %w", err)
+	}
+
+	// Walk every entry looking for TCP listen sockets on the old port
+	remapped := 0
+	for _, entry := range img.Entries {
+		fileEntry, ok := entry.Message.(*fdinfo.FileEntry)
+		if !ok {
+			continue
+		}
+		if fileEntry.GetType() != fdinfo.FdTypes_INETSK {
+			continue
+		}
+		isk := fileEntry.GetIsk()
+		if isk == nil {
+			continue
+		}
+		if isk.GetState() == tcpListenState && isk.GetSrcPort() == oldPort {
+			np := newPort
+			isk.SrcPort = &np
+			remapped++
+		}
+	}
+
+	if remapped == 0 {
+		return nil, nil, fmt.Errorf("no TCP listen sockets found with source port %d", oldPort)
+	}
+
+	// Encode the modified image to another temp file
+	tmpOut, err := os.CreateTemp("", "files-img-out-*.img")
+	if err != nil {
+		return nil, nil, fmt.Errorf("creating temp output file: %w", err)
+	}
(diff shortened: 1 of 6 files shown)


ImranullahKhann:
I have tested the functionality on Podman and CRI-O checkpoints. I will start working on writing the tests and would appreciate any feedback and guidance.

adrianreber:
Please take a look at providing unit tests and end to end tests.

ImranullahKhann:
@adrianreber, please have a look

adrianreber:
Try to add more unit tests. Especially the error branches are not covered well.

ImranullahKhann:
Please have a look

internal/archive_modifiers.go
```diff
@@ -0,0 +1,237 @@
...
+		return
+	}
+
+	for _, mapping := range mappingsSlice {
+		mappingMap, ok := mapping.(map[string]any)
+		if !ok {
+			continue
+		}
+
+		containerPort, ok := mappingMap["container_port"]
```
behouba, line 180: If i understand correctly, this match on the container port only, so `--tcp-listen-remap` could rewrite UDP mapping that share the same port (common situation with containers like [pihole](https://hub.docker.com/r/pihole/pihole) and [coredns](https://hub.docker.com/r/coredns/coredns)). Would it make sense to skip non TCP ports here?

```go
if proto, _ := mappingMap["protocol"].(string); proto != "" && proto != "tcp" {
    continue
}
```

internal/edit_archive.go
```diff
@@ -0,0 +1,235 @@
...
+	archiveInfo, err := archiveFile.Stat()
+	if err != nil {
+		return fmt.Errorf("failed to stat archive: %w", err)
+	}
+
+	// Detect Compression
+	b := make([]byte, 10)
+	n, err := io.ReadFull(archiveFile, b)
+	if err != nil {
+		return fmt.Errorf("failed to read archive magic bytes")
```
behouba, line 104:
```suggestion
		return fmt.Errorf("failed to read archive magic bytes: %w", err)
```

internal/edit_archive.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	comp := archive.DetectCompression(b[:n])
+	// Seek back so DecompressStream can read from the start
+	if _, err := archiveFile.Seek(0, 0); err != nil {
+		return fmt.Errorf("failed to seek archive")
```
behouba, line 109:
```suggestion
		return fmt.Errorf("failed to seek archive: %w", err)
```

internal/edit_archive.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	// Decompress the archive into a plan tar stream
+	tarStream, err := archive.DecompressStream(archiveFile)
+	if err != nil {
+		return fmt.Errorf("failed to decompress archive")
```
behouba, line 115:
```suggestion
		return fmt.Errorf("failed to decompress archive: %w", err)
```

internal/edit_archive.go
```diff
@@ -0,0 +1,235 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to create output file: %w", err)
+	}
+	outputPath := outFile.Name()
+
+	compressor, err := archive.CompressStream(outFile, comp)
+	if err != nil {
+		outFile.Close()
+		os.Remove(outputPath)
+		return fmt.Errorf("failed to create compressor")
```
behouba, line 130:
```suggestion
		return fmt.Errorf("failed to create compressor: %w", err)
```

internal/archive_modifiers.go
```diff
@@ -0,0 +1,237 @@
...
+	data, err := io.ReadAll(content)
+	if err != nil {
+		return nil, nil, fmt.Errorf("reading config.dump: %w", err)
+	}
+
+	// Parse into a generic map to preserve all fields
+	var config map[string]any
+	if err := json.Unmarshal(data, &config); err != nil {
+		return nil, nil, fmt.Errorf("parsing config.dump JSON: %w", err)
+	}
```
behouba, line 112: Big integers get rounded  here because of `map[string]any` will decode all numbers as float64. This can corrupt the output when the default system resource limits are overwritten ( `--ulimit`). 

Maybe something like this can keep the fidelity is unchanged.

```go
      dec := json.NewDecoder(bytes.NewReader(data))
      dec.UseNumber()
      var config map[string]any
      if err := dec.Decode(&config); err != nil {
              return nil, nil, fmt.Errorf("parsing config.dump JSON: %w", err)
      }
```

internal/archive_modifiers.go
```diff
@@ -0,0 +1,237 @@
...
+func remapSpecDump(hdr *tar.Header, content io.Reader, oldPort, newPort string) (*tar.Header, []byte, error) {
+	data, err := io.ReadAll(content)
+	if err != nil {
+		return nil, nil, fmt.Errorf("reading spec.dump: %w", err)
+	}
+
+	var spec map[string]any
+	if err := json.Unmarshal(data, &spec); err != nil {
+		return nil, nil, fmt.Errorf("parsing spec.dump JSON: %w", err)
+	}
```
behouba, line 137: same possible issue as above.

behouba (commented):
@ImranullahKhann thank you for working on this. I added some comments, feel free to take a look when you can.

PR #210, opened 2026-03-26
add missing diff socket logic

hxadogar:
diff `--sockets` was registered but not actually any compared logic.
now shows sockets added or removed.
it handles all socket types.

i have tested and verified manually.
and i will try to follow up with a dedicated pr for tests.


adrianreber:
Please add the tests to this PR as a second commit.

hxadogar:
of course, will add tests shortly.

for now here is my manual testing output:

```
➜ sudo ./checkpointctl diff socket-8080.tar socket-8081_8082.tar --sockets
╔════════════════════════════════════════════════════════════════╗
║ Checkpoint Diff                                                ║
╠════════════════════════════════════════════════════════════════╣
║ Container: socket-test                                         ║
║ Image:     localhost/socket-test:latest                        ║
║ ID:        2088c8c5543c036e2b706e2ca743cd41c86c81b03037a524... ║
╚════════════════════════════════════════════════════════════════╝

Checkpoint A:
  Created: 2026-03-26T16:53:05+05:00
  Size:    5551762 bytes

Checkpoint B:
  Created: 2026-03-26T16:53:05+05:00
  Size:    5551910 bytes

┌─ Memory Changes ─────────────────────────────────────────────┐
│ = No change
└──────────────────────────────────────────────────────────────┘
┌─ Process Changes ────────────────────────────────────────────┐
└──────────────────────────────────────────────────────────────┘
┌─ Socket Changes ─────────────────────────────────────────────┐
│ Added:
│   + PID 1     TCP     0.0.0.0:8081 -> 0.0.0.0:0
│   + PID 1     TCP     0.0.0.0:8082 -> 0.0.0.0:0
│ Removed:
│   - PID 1     TCP     0.0.0.0:8080 -> 0.0.0.0:0
└──────────────────────────────────────────────────────────────┘
Summary:
Checkpoint comparison for container socket-test
Sockets: +2 -1
```

hxadogar:
i have added bats tests and i think its ready for review

adrianreber:
Thanks. Also add a couple of unit tests, especially for the error cases which are not covered via bats tests.

hxadogar:
i added the unit tests and try to covered the error cases too (not too much)

you can take another look when you get time

adrianreber:
Looks good to me. @rst0git What do you think? The commit messages could contain some more details about the changes.

cmd/diff.go
```diff
@@ -241,6 +244,29 @@ type OpenFile struct {
 	Path string `json:"path"`
 }
 
+// Socket types (mirroring internal/json.go)
```
rst0git: Instead of duplicating these types, can we import them?

cmd/diff.go
```diff
@@ -564,6 +699,50 @@ func renderTreeDiff(result *DiffResult) {
...
+						socket.Source, socket.SrcPort,
+						socket.Dest, socket.DstPort)
+				default:
+					fmt.Printf("%s\n", truncate(socket.Address, 40))
+				}
+			}
+		}
+
+		if len(result.SocketChanges.Removed) > 0 {
+			fmt.Println("│ Removed:")
```
rst0git: This also appears to be copy-pasted. Try to avoid code duplication.

For example, you can create a helper function that contains the shared functionality:
```go
func formatSocketEndpoint(socket SocketInfo) string {
      switch socket.Type {
      case "TCP", "UDP":
              return fmt.Sprintf("%s:%d -> %s:%d",
                      socket.Source, socket.SrcPort,
                      socket.Dest, socket.DstPort)
      default:
              return truncate(socket.Address, 40)
      }
}
```
then simplify both loops to:

```Go
for _, socket := range result.SocketChanges.Added {
    fmt.Printf("│   + PID %-5d %-8s%s\n", socket.PID, socket.Protocol, formatSocketEndpoint(socket))
}

for _, socket := range result.SocketChanges.Removed {
    fmt.Printf("│   - PID %-5d %-8s%s\n", socket.PID, socket.Protocol, formatSocketEndpoint(socket))
}
```

hxadogar: both Done

cmd/diff.go
```diff
@@ -444,6 +471,82 @@ func compareFileDescriptors(fdsA, fdsB []FileDescriptorEntry) *FileDiff {
...
+func compareSockets(sksA, sksB []internal.SkNode) *SocketDiff {
+	diff := &SocketDiff{}
+
+	// Build socket maps
+	// Key format: PID:Protocol:Type:Address:SrcPort:Dest:DstPort
+	mapA := make(map[string]SocketInfo)
+	mapB := make(map[string]SocketInfo)
+
+	for _, entry := range sksA {
+		for _, socket := range entry.OpenSockets {
```
hxadogar, line 497: should i refactor this too with buildMap helper (could use anonymous func to keep it contained)
or
leave it as-is since it follows the same pattern as compareFileDescriptors()?

adrianreber:
@hxadogar Sorry, I pressed the wrong button. Can you remove my merge commit? I can also do it if you want.

rst0git:
@adrianreber No worries! I will update the PR soon

hxadogar:
Thanks @rst0git! I was about to handle it

rst0git:
@hxadogar I've revised the patches in this branch https://github.com/rst0git/checkpointctl/tree/socket-diff
Would you be able to confirm if you are happy with the changes?

rst0git:
I will go ahead and merge these changes; we can make any other changes in follow-up PRs

PR #214, opened 2026-04-19
ci: install git in the Fedora container

adrianreber:
Especially as codecov seems to require it now.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -12,14 +12,16 @@ jobs:
       image: fedora:latest
       options: --privileged
     steps:
+      - name: Install tools
+        run: |
+          dnf install -y bats iptables iproute jq criu golang make git
+      - name: Fix git safe directory
+        run: git config --global --add safe.directory /__w/checkpointctl/checkpointctl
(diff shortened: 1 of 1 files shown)


.github/workflows/coverage.yml
```diff
@@ -12,14 +12,16 @@ jobs:
       image: fedora:latest
       options: --privileged
     steps:
+      - name: Install tools
+        run: |
+          dnf install -y bats iptables iproute jq criu golang make git
+      - name: Fix git safe directory
+        run: git config --global --add safe.directory /__w/checkpointctl/checkpointctl
```
rst0git, line 19: Would it work if we use `"$GITHUB_WORKSPACE"` here (https://github.com/actions/checkout/issues/2031#issuecomment-2753735800)?

adrianreber: Yes, I was using that also, but I wanted to include this also in the premade containers in the other script and that would probably not work there. I can still change it. This is more a quick fix to make sure CI works again here. Let me know if you prefer to have the variable here (even if it is changed with the container setup).

PR #217, opened 2026-04-20
ci: use pre-built CI container in coverage and test workflows

adrianreber:
Replace fedora:latest with the weekly built CI container image from ghcr.io and remove the dnf install steps, as all packages are already included in the container.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -4,17 +4,18 @@ on: [push, pull_request]
 
 permissions:
   contents: read
+  packages: read
 
 jobs:
   coverage:
     runs-on: ubuntu-latest
     container:
-      image: fedora:latest
+      image: ghcr.io/${{ github.repository }}/ci:latest
+      credentials:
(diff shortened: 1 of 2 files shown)


adrianreber:
Package are not public:

```
 Public

Make this package visible to anyone.
Setting is disabled by organization administrators. 
```

and I cannot change it.

rst0git:
@adrianreber We can also use [`quay.io/criu/checkpointctl-ci`](https://quay.io/repository/criu/checkpointctl-ci). I've added username and secret to the checkpointctl repository.

adrianreber:
Thanks. It would be more efficient to use the ghcr registry as it is much closer and faster. We need an organization admin to allow public images. I guess Andrei should be able to do it.

adrianreber:
Wait, better idea coming.

rst0git:
>It would be more efficient to use the ghcr registry as it is much closer and faster. 

btw, I'm not sure if it is actually faster: https://github.com/orgs/community/discussions/27080

PR #227, opened 2026-09-06
docs: add diff man page and update docs Makefile

srinivasr:
Add a man page for the `diff` subcommand (`docs/checkpointctl-diff.adoc`) and include it in `docs/checkpointctl.adoc`.

Also update `docs/Makefile` so that `build`, `diff`, `list`, and `plugin` man pages are included in `SRC1` and get installed by `make install`.

In addition, fix a copy-paste error in `checkpointctl-inspect.adoc` where the help text referenced `memparse`, document its `--metadata` flag, and align section formatting in `checkpointctl-build.adoc`.


rst0git:
@srinivasr Thank you for working on this!

I've updated the pull request to split the changes into logical commits (see [contributor guidelines](https://github.com/checkpoint-restore/checkpointctl/blob/main/CONTRIBUTING.md)) and added a patch to fix the diff command so that the functionality with `--ps-tree-env` matches the intended behaviour.

rst0git (approved):

PR #229, opened 2026-09-22
internal: clean up temporary directories on task creation failure

srinivasr:
CreateTasks creates one os.MkdirTemp directory per archive, but every error return between the first successful extraction and the end of the loop leaves the directories already created behind:

go
if !checkpointDirExists {
    return nil, fmt.Errorf("checkpoint directory is missing in the archive file: %s", input)
}

To reproduce: run a command over several archives where one after the first fails, e.g. checkpointctl show valid.tar broken.tar where broken.tar has no checkpoint/ directory. The command errors out and the checkpointctl* directory created for valid.tar stays in $TMPDIR. Every failed run leaks one directory per successfully processed archive.

Fix: call CleanupTasks on the tasks collected so far before each error return, and remove the new directory itself when UntarFiles fails on it.

Tests: two new cases in internal/utils_test.go - a missing checkpoint directory on the second archive, and an UntarFiles failure from a file/directory name conflict - both asserting no checkpointctl* entries remain in TMPDIR. Error paths that need injected filesystem failures (MkdirTemp itself failing, RemoveAll errors) are not covered.


behouba (approved):
Thank you @srinivasr. LGTM!  Maybe a deferred call of `CleanupTasks` would centralize the cleanup and help cover future early returns paths.

srinivasr:
Good call, done in 275e286. One defer at the top handles all the error returns now, moved the append before UntarFiles so the partial dir gets cleaned too. Went with a success flag instead of named returns since err gets shadowed in the loop and the defer would have silently done nothing. Tests untouched.

